PR #12, opened 2019-10-14
Fix overeager indenting with assignments

dansanduleac:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==


ferozco (commented):
Lets ship the assignment changes, since it actually improves user code and then add the test cases in a separate PR

dansanduleac:
Don't think this is necessary anymore

PR #18, opened 2019-10-14
Improvement: IntelliJ plugin with configurable implementation

dansanduleac:
## Before this PR

The idea-plugin just bundles the current version of the formatter.

However, the plugin version is global to an IntelliJ installation, so all projects would be forced to use that version of the formatter. This is bad because IntelliJ formatting could get out of sync with `./gradlew format`

## After this PR
==COMMIT_MSG==
IntelliJ plugin can be automatically configured with a custom version of the formatter, by applying `com.palantir.java-format` in gradle and setting the `palantirJavaFormat.implementationVersion`.
==COMMIT_MSG==

In order to achieve this, we've separated palantir-java-format-api out of the formatter. This is a super thin API layer, and the implementation is found by service loading it out of the implementation jar.


ferozco (approved):
small nits then should be good to go! super excited about this :)

ferozco (approved):
small nits then should be good to go! super excited about this :)

ferozco (approved):
small nits then should be good to go! super excited about this :)

iamdanfox (approved):

PR #44, opened 2019-10-21
Improvement: better inlining of method reference

dansanduleac:
## Before this PR

The formatter would produce code that fails checkstyle, by indenting a block before a method reference too much.

## After this PR
==COMMIT_MSG==
Prefer inlining method references after whatever expression they follow.
==COMMIT_MSG==


iamdanfox (approved):

dansanduleac:
come on bulldozer, you can do it

PR #45, opened 2019-10-21
Fix: NON-NLS comments move before `+` in string concatenations

dansanduleac:
## Before this PR

NON-NLS comments after `+` in chains of string concatenations are moved such that they don't follow the original string anymore.

## After this PR
==COMMIT_MSG==
NON-NLS comments are moved behind `+` tokens in string concatenations. This is a heuristic that doesn't look at what the thing before the `+` is, so it might produce confusing results.
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/OpsBuilder.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/OpsBuilder.java
@@ -541,21 +541,32 @@ public final class OpsBuilder {
                         tokOps.put(j, SPACE);
                     }
                     // Now we've seen the Token; output the toksAfter.
+
+                    // Reordering of NON-NLS comments that might follow a `+` in a chain of string concatenations, in
+                    // order to move the comments before the Break that precedes the `+` token.
+                    Op previousOp = ops.get(i - 1);
+                    boolean nonNlsCommentsAfterPlus = token.getToksAfter().stream()
+                                    .anyMatch(OpsBuilder::isNonNlsComment)
+                            && token.getTok().getText().equals("+")
+                            && previousOp instanceof Break;
+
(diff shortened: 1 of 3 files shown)


palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/NON-NLS.output
```diff
@@ -3,15 +3,11 @@ package test;
...
-            "field_2,"
-            + //$NON-NLS-1$
-            "field_3,"
-            + //$NON-NLS-1$
-            "field_4"; //$NON-NLS-1$
+    private static final String MULTIPLE_LINE_NON_NLS = "field_0," //$NON-NLS-1$
+            + "field_1," //$NON-NLS-1$
+            + "field_2," //$NON-NLS-1$
+            + "field_3," //$NON-NLS-1$
+            + "field_4"; //$NON-NLS-1$
```
carterkozak, line 10: ❤️ 😍

palantir-java-format/src/main/java/com/palantir/javaformat/OpsBuilder.java
```diff
@@ -620,6 +631,10 @@ private static int getI(Input.Token token) {
         return newOps.build();
     }
 
+    private static boolean isNonNlsComment(Input.Tok tokAfter) {
+        return tokAfter.isSlashSlashComment() && tokAfter.getText().contains("$NON-NLS");
```
carterkozak, line 635: May be able to make this more specific, nls comments are required to start with `//$NON-NLS-`, even subsequent numbers, e.g.
```java
String f() {
  return "a" + "b"; //$NON-NLS-1$ //$NON-NLS-2$
}
```

Realistically I'm not sure that makes much difference.

carterkozak (approved):

PR #46, opened 2019-10-21
NON-NLS more difficult example

carterkozak:
==COMMIT_MSG==
NON-NLS more difficult example
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/NON-NLS.input
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/NON-NLS.input
@@ -17,4 +17,6 @@ class MultilineStringConstant {
                     "field_3," +
                     "field_4";
 
+    private static final String SINGLE_LINE_NON_NLS = "field_0," + "field_1," + "field_2," +  "field_3," + "field_4"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+
 }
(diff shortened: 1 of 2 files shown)


iamdanfox:
Ah dang :/

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/NON-NLS.input
```diff
@@ -17,4 +17,6 @@ class MultilineStringConstant {
                     "field_3," +
                     "field_4";
 
+    private static final String SINGLE_LINE_NON_NLS = "field_0," + "field_1," + "field_2," +  "field_3," + "field_4"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
```
dansanduleac, line 20: Is this for real??
I'd like to have a look at a comprehensive set of rules showing all the ways that these comments work.

* do we need an extra `//` for each `$NON-NLS` instance?
* does each comment map to a string literal, or could they be mapped to string constants / variables as well?
* do we only need to handle string concatenations or other AST shapes too?

carterkozak: 1. Yep, that's correct
2. Each comment is associated with a string literal, these are string constants in source
3. This applies beyond concatenation, for example `return getFoo("a", "b") //$NON-NLS-1$ //$NON-NLS-2$` could be rewritten to
```java
return getFoo(
    "a", //$NON-NLS-1$
    // note that the index is updated from 2 to 1 after a line split
    "b") //$NON-NLS-1$
```

dansanduleac: Ok. I think for the sake of simplicity, we're going to have to force the comments to come right after each string literal + random other language tokens that are attached to it unbreakably (like `,;)`), and always give them a `$NON-NLS-1$` tag as a result. Otherwise we can't be sure, at the level where we process this, that the lines won't be broken later.

How does that sound?

carterkozak: Sounds reasonable to me :-)

PR #47, opened 2019-10-22
Break: rename api->spi, remove JavaFormatterOptions and expose new method

dansanduleac:
## Before this PR

Baseline is reaching into the implementation to call `formatSourceAndFixImports` in order to get string reflowing.

## After this PR
==COMMIT_MSG==
* Rename api -> spi.
* Remove JavaFormatterOptions from the SPI.
* Expose `formatSourceReflowStringsAndFixImports` from the API, which formats a whole file returning the formatted file as a String.
==COMMIT_MSG==


palantir-java-format-api/src/main/java/com/palantir/javaformat/java/FormatterService.java
```diff
@@ -18,6 +18,19 @@
...
+     * <p>Fixing imports includes ordering, spacing, and removal of unused import statements.
+     *
+     * @param input the input string
+     * @return the output string
+     * @throws FormatterException if the input string cannot be parsed
+     * @see <a href="https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing">Google Java
+     *     Style Guide - 3.3.3 Import ordering and spacing</a>
+     */
+    String formatSourceReflowStringsAndFixImports(String input) throws FormatterException;
 }
```
iamdanfox: could we also rename this project from -api to -spi to emphasise it's a service provider interface??

iamdanfox (approved):

PR #48, opened 2019-10-22
Standalone gradle plugins

iamdanfox:
## Before this PR

We're spreading logic between baseline and this repo, and don't actually have a particularly simple usage story (we currently tell people to apply `com.palantir.baseline-format` and add `com.palantir.baseline-format.palantir-java-format` to their `gradle.properties`)

## After this PR
==COMMIT_MSG==
The standalone `com.palantir.java-format` plugin provides a `formatDiff` task. IntelliJ setup has been moved to a `com.palantir.java-format-idea` plugin.
==COMMIT_MSG==

This should hopefully reduce the coupling between baseline & this repo and make it easier for people to just try out our formatter if they want to.

## Possible downsides?

- this is diverging from spotless slightly, perhaps an alternative would be to get `palantirJavaFormat()` merged into spotless upstream??


dansanduleac:
👍

PR #49, opened 2019-10-22
Update sinceBuild, don't publish IDEA plugin unless clean tag

dansanduleac:
## Before this PR

Keep getting compatibility error emails due to some deprecated methods being used in old versions of IDEA.

* `CodeStyleManager.fillIndent()`
* `CodeStyleManager.getIndent()`
* `CodeStyleManager.isLineToBeIndented()`
* `CodeStyleManager.zeroIndent()`

See e.g. https://plugins.jetbrains.com/plugin/13180-palantir-java-format/update/70978

These methods are still accessible from the old class, but have since been moved to a super class.

## After this PR
==COMMIT_MSG==
Bump minimum intellij build to 182 to get rid of spurious deprecation warnings on every publish.
==COMMIT_MSG==

### DIFF
--- a/idea-plugin/build.gradle
+++ b/idea-plugin/build.gradle
@@ -28,17 +28,14 @@ intellij {
 patchPluginXml {
     pluginDescription = "Formats source code using the palantir-java-format tool."
     version = project.version
-    sinceBuild = '173' // TODO: test against this version of IntelliJ to ensure no regressions
+    sinceBuild = '182' // TODO: test against this version of IntelliJ to ensure no regressions
     untilBuild = ''
 }
 
 publishPlugin {
     token = System.env.JETBRAINS_PLUGIN_REPO_TOKEN
-    if (!versionDetails().isCleanTag) {
-        channels 'eap'
-    }
 }
-
+tasks.publishPlugin.onlyIf { versionDetails().isCleanTag }
 tasks.publish.dependsOn publishPlugin
 
 configurations {


idea-plugin/build.gradle
```diff
@@ -28,17 +28,14 @@ intellij {
 patchPluginXml {
```
ferozco, line 28: should we also bump the version of intellij we test with?

idea-plugin/build.gradle
```diff
@@ -28,17 +28,14 @@ intellij {
...
-    sinceBuild = '173' // TODO: test against this version of IntelliJ to ensure no regressions
+    sinceBuild = '182' // TODO: test against this version of IntelliJ to ensure no regressions
     untilBuild = ''
 }
 
 publishPlugin {
     token = System.env.JETBRAINS_PLUGIN_REPO_TOKEN
-    if (!versionDetails().isCleanTag) {
-        channels 'eap'
-    }
```
dansanduleac, line 39: this was pointless, and I'm not even sure how I could consume the snapshots if I wanted to (even tho i'm subscribed to the EAP plugin list)

ferozco:
👍

idea-plugin/build.gradle
dansanduleac: We can do that later.
I know for a fact it doesn't work fully well in 193.4099+ and keeps throwing exceptions, but need to figure out why and how to fix. Funnily enough their automatic detection hasn't picked that up.

PR #54, opened 2019-10-24
State is now immutable

dansanduleac:
## Before this PR

While formatting a Doc via `Doc.computeBreaks`, we relied on a lot of state being kept around in the Docs themselves (e.g. in Break, Tok, Level, BreakTag).

This doesn't interact well with branching as done for levels with BreakBehavior = `preferBreakingLastInnerLevel` (where it's allowed to either inline a Level, or break it) as we then need to redo work upon a branch prediction failure, in order to ensure the end state is correct.

## After this PR
==COMMIT_MSG==
State is now immutable and keep track of all the formatting outcomes (what levels were broken, what breaks were taken, how comments were split) in fast-update immutable data structures.
==COMMIT_MSG==

## Possible downsides?
* introduce new library functionaljava in order to take advantage of fast-update immutable data structures that don't exist by the JRE or guava.


palantir-java-format/src/main/java/com/palantir/javaformat/Output.java
```diff
@@ -16,33 +16,16 @@
...
-            taken = Optional.of(broken);
-        }
-
-        public void reset() {
-            taken = Optional.empty();
-        }
-
-        public boolean wasBreakTaken() {
-            return taken.orElse(false);
-        }
```
iamdanfox, line 44: :+1: v nice that this has all coalesced into the State class!

palantir-java-format/src/main/java/com/palantir/javaformat/doc/HasUniqueId.java
```diff
@@ -0,0 +1,14 @@
+package com.palantir.javaformat.doc;
+
+import fj.Ord;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public abstract class HasUniqueId {
```
iamdanfox: would be nice to tidy up their package structure a little bit so this doens't have to be public!

iamdanfox: Also worth a little comment saying this is just because we need a _stable_ order and functionaljava uses redblack trees not linkedhashset/maps?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/BreakTag.java
```diff
@@ -0,0 +1,7 @@
+package com.palantir.javaformat.doc;
+
+import com.google.errorprone.annotations.Immutable;
+
+/** Unique identifier for a break. */
```
iamdanfox: could we @link to ConditionalBlankLine etc to emphasise why we'd care about having a "unique identifier for a break"?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/HasUniqueId.java
```diff
@@ -0,0 +1,14 @@
+package com.palantir.javaformat.doc;
+
+import fj.Ord;
+import java.util.concurrent.atomic.AtomicInteger;
+
+abstract class HasUniqueId {
```
iamdanfox, line 10: little javadoc to justify this things existence?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -40,7 +40,7 @@
      * How many branches we are allowed to take (i.e how many times we can consider breaking vs not breaking the current
      * level) before we stop branching and always break, which is the google-java-format default behaviour.
      */
-    private static final int MAX_BRANCHING_COEFFICIENT = 7;
+    private static final int MAX_BRANCHING_COEFFICIENT = 20;
```
iamdanfox, line 46: @dansanduleac explained in person this can be dialled up (and still be performant) because the formatter was not previously counting branching accurately (so was actually doing lots more work than a max of 7 would have allowed!)

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -51,13 +51,6 @@
     private final Optional<String> debugName;
     private final List<Doc> docs = new ArrayList<>(); // The elements of the level.
 
-    // State that needs to be preserved between calculating breaks and
-    // writing output.
-    // TODO(cushon): represent phases as separate immutable data.
```
iamdanfox, line 56: 🤩

palantir-java-format/src/main/java/com/palantir/javaformat/doc/State.java
```diff
@@ -16,93 +16,204 @@
...
-    final int branchingCoefficient;
+    public abstract int branchingCoefficient();
+
+    protected abstract Set<BreakTag> breakTagsTaken();
+
+    protected abstract TreeMap<Break, BreakState> breakStates();
+
+    protected abstract TreeMap<Level, LevelState> levelStates();
+
+    protected abstract List<P2<Tok, TokState>> tokStates();
```
iamdanfox: wanna justify why this isn't a TreeMap too?

versions.props
```diff
@@ -12,6 +12,7 @@ gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle-idea-ext = 0.5
 junit:junit = 4.12
 org.assertj:assertj-core = 3.13.2
 org.derive4j:* = 1.1.1
+org.functionaljava:functionaljava = 4.8
```
iamdanfox, line 15: Just for posterity's sake - what was the specific performance slowness from using plain guava ImmutableLists and ImmutableMaps that motivated you to switch to this?

dansanduleac: We keep track of state associations for every Level, Break and Tok, of which there are many. Therefore the performance goes from `N -> N^2` in the number of levels + toks + breaks (of which Toks are dominant).
When I tried to use guava immutables, it made the super degenerate test case (B24909927) take about 130x the time before this PR.

palantir-java-format/src/main/java/com/palantir/javaformat/doc/State.java
dansanduleac: I thought this is a perf optimization since they are all first written and then only the final state is read as a map, but actually these are only comments, so there aren't that many.
Therefore, I've made this a TreeMap too.

iamdanfox (approved):

dansanduleac:
👍

PR #70, opened 2019-11-12
Limit the max width of method chain dots

### LINKED ISSUE #69: Hard to read long method chains
## What happened?

https://github.com/palantir/conjure-java-runtime/pull/1285/files#diff-3d840aee82ecd3ad6d02f77a1bdd0239L56

Human-authored:

```java
return new Response.Builder()    
        .request(request)    
        .protocol(Protocol.HTTP_1_1)    
        .code(code)    
        .message("unused")    
        .build();
```

PJF:

```java
return new Response.Builder().request(request).protocol(Protocol.HTTP_1_1).code(code).message("unused").build();
```

Don't allow method chains to extend too far to the right.

We should probably limit the absolute column past which method chain `.`s are not allowed anymore to something more sensible than 120, like 80.

dansanduleac:
## After this PR
==COMMIT_MSG==
Limit how far dots may appear in long method chains to 80 chars.

Note that this doesn't currently apply to prefixes (such as `foo.bar().baz().stream()`) because prefixes are also used to group together fully qualified class names and it's a bit trickier to handle that case.

Fixes #69.
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
@@ -69,7 +69,7 @@ public abstract class OpenOp implements Op {
 
     @Override
     public void add(DocBuilder builder) {
-        builder.open(plusIndent(), breakBehaviour(), breakabilityIfLastLevel(), debugName());
+        builder.open(this);
     }
 
     public static Builder builder() {
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
@@ -28,13 +28,19 @@ import org.immutables.value.Value.Default;
  */
 @Value.Immutable
 public abstract class OpenOp implements Op {
+    /** The extra indent inside this level. */
     public abstract Indent plusIndent();
 
+    /**
+     * When this level doesn't fit on one line, controls whether this level is to be broken (its breaks taken) or
+     * partially inlined onto the current line.
+     */
     @Default
     public BreakBehaviour breakBehaviour() {
         return BreakBehaviours.breakThisLevel();
     }
 
+    /** If it's the last level of its parent, when to inline this level rather than break the parent. */
     @Default
     public LastLevelBreakability breakabilityIfLastLevel() {
         return LastLevelBreakability.ABORT;
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
@@ -18,6 +18,7 @@ import com.palantir.javaformat.doc.Doc;
 import com.palantir.javaformat.doc.DocBuilder;
 import com.palantir.javaformat.doc.Level;
 import java.util.Optional;
+import java.util.OptionalInt;
 import org.immutables.value.Value;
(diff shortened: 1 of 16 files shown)


palantir-java-format/src/main/java/com/palantir/javaformat/OpenOp.java
```diff
@@ -69,7 +79,7 @@ public static Op make(
 
     @Override
     public void add(DocBuilder builder) {
-        builder.open(plusIndent(), breakBehaviour(), breakabilityIfLastLevel(), debugName());
+        builder.open(this);
```
dansanduleac, line 82: I've also taken the opportunity of passing through OpenOp right into Level rather than unpacking its properties through multiple methods, as it was getting laborious to add new fields.

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Break.java
```diff
@@ -25,25 +24,19 @@
...
-    private final Optional<BreakTag> optTag;
-
-    private Break(FillMode fillMode, String flat, Indent plusIndent, Optional<BreakTag> optTag) {
-        this.fillMode = fillMode;
-        this.flat = flat;
-        this.plusIndent = plusIndent;
-        this.optTag = optTag;
-    }
+@Value.Immutable
+public abstract class Break extends Doc implements Op {
```
dansanduleac, line 32: I converted this to an immutable thinking I wanted to add new fields to it but ended up not doing that.
Still, I prefer this style and it makes it easier to change in the future.

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Break.java
```diff
@@ -143,17 +131,13 @@ public void write(State state, Output output) {
             output.append(state, "\n", EMPTY_RANGE);
             output.indent(breakState.newIndent());
         } else {
-            output.append(state, flat, range());
+            output.append(state, flat(), range());
         }
     }
 
-    @Override
-    public String toString() {
```
ferozco, line 151: nice

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -122,17 +102,46 @@ String computeFlat() {
 
     @Override
     public State computeBreaks(CommentsHelper commentsHelper, int maxWidth, State state) {
-        float thisWidth = getWidth();
-        if (state.column() + thisWidth <= maxWidth) {
-            return state.withColumn(state.column() + (int) thisWidth)
-                    .withLevelState(this, ImmutableLevelState.of(true));
+        Optional<State> oneLine = tryToFitOnOneLine(maxWidth, state);
```
ferozco:
```suggestion
        return tryToFitOnOneLine(maxWidth, state)
            .orElseGet(() -> state.updateAfterLevel(getBreakBehaviour().match(new BreakImpl(commentsHelper, maxWidth, state));
```

ferozco (commented):
LGTM

ferozco:
👍

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
dansanduleac: yep looks better, tho I kept the orElse bit on two lines as I think it looks cleaner

PR #71, opened 2019-11-12
Fix: Limit dotted chains for expressions starting with complex expression

dansanduleac:
## Before this PR

#70 didn't address all cases of method chains.

Specifically, cases where the dotted chain started with a constructor or a complex expression rather than a field / method call.

## After this PR
==COMMIT_MSG==
Limit dotted chains for expressions starting with a complex expression such as a parenthesized expression or a constructor. 
==COMMIT_MSG==


dansanduleac:
Verified this makes positive changes in the conjure-java-runtime PR.

ferozco (approved):

PR #72, opened 2019-11-13
Move input metadata collection out of Output

dansanduleac:
## Before this PR

`Output` (and `JavaOutput`) acted as a builder for and kept track of input metadata, such as what newlines to preserve and what ranges are partially formattable. 

This information was computed once in OpsBuilder and read only when writing to the JavaOutput.

However, this arrangement means you can't easily create multiple JavaOutputs (if you want to see the what a current formatting path produced so far).

## After this PR
==COMMIT_MSG==
Move input metadata collection out to separate class that is then passed into JavaOutput.

This enables creating multiple JavaOutputs for debugging purposes, without losing this metadata that was already fully known before the last stage of formatting (`Doc.computeBreaks`).
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaOutput.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaOutput.java
@@ -352,7 +352,7 @@ public final class JavaOutput extends Output {
                 .add("lastK", lastK)
                 .add("spacesPending", spacesPending)
                 .add("newlinesPending", newlinesPending)
-                .add("blankLines", inputMetadata.blankLines())
+                .add("inputMetadata", inputMetadata)
                 .add("super", super.toString())
                 .toString();
     }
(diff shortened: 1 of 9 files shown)


palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaOutput.java
```diff
@@ -193,11 +172,6 @@ void flush() {
 
     // The following methods can be used after the Output has been built.
 
-    @Override
-    public CommentsHelper getCommentsHelper() {
```
dansanduleac, line 197: This was not used at all in this class, was only read from the Formatter.format method so it was essentially used to bundle two values into one.

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaOutput.java
```diff
@@ -374,7 +352,7 @@ public String toString() {
                 .add("lastK", lastK)
                 .add("spacesPending", spacesPending)
                 .add("newlinesPending", newlinesPending)
-                .add("blankLines", blankLines)
+                .add("blankLines", inputMetadata.blankLines())
```
iamdanfox: just plain inputMetadata?

iamdanfox (approved):

dansanduleac:
🐛

PR #73, opened 2019-11-13
HTML debugging view

iamdanfox:
## Before this PR

I found it hard to imagine how all of the abstractions in this project fit together, which made iterating on the actual logic a bit confusing.

## After this PR

==COMMIT_MSG==
A HTML debugging view can be generated to help iterate on the formatter logic
==COMMIT_MSG==

Shows how `Op`s are laid out, before building them into `Levels`:

image

Shows the different paths the formatter took to reach the output, and allows exploring what the various "explorations" from each level would have produced if selected:

image

formatter-exploration-2

Before merging:
- [x] don't run this during normal formatting


palantir-java-format/build.gradle
```diff
@@ -10,6 +10,9 @@ dependencies {
     api 'com.google.guava:guava'
     implementation 'com.google.errorprone:javac-shaded'
     implementation 'org.functionaljava:functionaljava'
+    implementation 'com.fasterxml.jackson.core:jackson-databind' // purely for the debugger
+    implementation 'com.fasterxml.jackson.datatype:jackson-datatype-jdk8'
+    implementation 'com.fasterxml.jackson.module:jackson-module-parameter-names'
```
iamdanfox, line 15: @dansanduleac I think I'd like to keep the dependencies of palantir-java-format as minimal as possible - thoughts on keeping these in the `:debugger` project and taking a compileOnly dependency on that project?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Obs.java
```diff
@@ -0,0 +1,165 @@
...
+ */
+
+package com.palantir.javaformat.doc;
+
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.function.Function;
+import javax.annotation.CheckReturnValue;
+
+public interface Obs {
```
iamdanfox: Think we gotta tell people a bit of a story about what's going on here.

ferozco:
@dansanduleac could you push up a gif of your latest work? Very interested to see this in action

dansanduleac:
👍

iamdanfox:
:+1:

PR #77, opened 2019-11-14
Minor internal renames

iamdanfox:
As a pre-requisite to the debug UI work, I found these class names really misleading (especially because we had Token, Token and Tok).


dansanduleac (approved):

iamdanfox:
:+1:

PR #80, opened 2019-11-19
Fix: reformat on commit uses PJF

dansanduleac:
## Before this PR
Ticking "Before commit -> Reformat code" when committing from within IntelliJ would enter a code path that skips the PJF formatter and uses the built-in IntelliJ one instead.

This code path would call `CodeStyleManager.reformatTextWithContext(PsiFile, ChangedRangesInfo)` which the `CodeStyleManagerDecorator` would just delegate to the default implementation.
Currently, PalantirCodeStyleManager only overrides one of the two "reformatTextWithContext" methods.

## After this PR
==COMMIT_MSG==
"Before commit -> Reformat code" now correctly uses the PJF formatter if it is enabled, rather than the IntelliJ default formatter.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-80.v2.yml
+++ b/changelog/@unreleased/pr-80.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: '"Before commit -> Reformat code" now correctly uses the PJF formatter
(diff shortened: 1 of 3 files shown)


ferozco:
👍

PR #94, opened 2019-12-04
Don't always break at method calls with prefix

dansanduleac:
## Before this PR

We were allowing the inlining logic to stop (accept an inlining) whenever the last "level" (in most cases, argument to a method call) was a dotted construction (like a method call chain) that contains a prefix which PJF would want to keep on one line (`SafeArg.of` being the most prevalent case of this).

This would cause degenerate formattings like:

```java
Preconditions.checkArgument(
        hostPortString.charAt(0) == '[', "Bracketed host-port string must start with a bracket", SafeArg.of(
                "port", hostPortString));
```

## After this PR
==COMMIT_MSG==
Stop indiscriminately stopping an inlining at method calls like `SafeArg.of`.
==COMMIT_MSG==


iamdanfox:
Shall we do a preview on c-j-r and see what we get??

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-long-safearg.output
```diff
@@ -0,0 +1,8 @@
+class PalantirLongSafeArg {
+    public static void foo() {
+        Preconditions.checkArgument(
+                hostPortString.charAt(0) == '[',
+                "Bracketed host-port string must start with a bracket",
+                SafeArg.of("port", hostPortString));
```
iamdanfox, line 6: 🤩

dansanduleac:
Example changes compared to 0.3.2: 
https://github.com/palantir/conjure-java-runtime/pull/1354/commits/b5d6f00b8aeb1e9b0e4edc234c7a0cac89bf5c18

iamdanfox (approved):

iamdanfox:
@dansanduleac would you mind picking this up in gradle-baseline and getting a release of that out too pls?

PR #99, opened 2019-12-06
Allow inlining method calls if current inlining is simple

dansanduleac:
## Before this PR

#63 fixed a degenerate case where `SafeArg.of(` and other similar `X.y(` method calls could end up with their parameters on a new line, even if this was just an argument that was part of a long method call with other arguments.

Unfortunately, this fix also broke a bunch of situations that were relying on this inlining logic, and whose code consequently blew up to many many levels of indentation, breaking the 120 character limit.

## After this PR
==COMMIT_MSG==
Bring back inlining the prefix of method calls *iff* the levels inlined so far have all been simple.

By simple, in practice we mean a single method call (rather than chained method calls), or a single method parameter.  This is defined as a property on the Level `OpenOp`s created in `JavaInputAstVisitor`.
==COMMIT_MSG==


iamdanfox (approved):

iamdanfox:
Would kinda like to have the specific test case from fforms that prompted this checked in!

dansanduleac:
🚂

PR #101, opened 2019-12-06
Don't allow inlined level to exceed columnLimitBeforeLastBreak

### LINKED ISSUE #100: Prevent long method chains even when inlining
## What happened?

Currently, we prevent one-lining a level with method chains where the last dot reaches past a column (80), as of #70.

However, we don't apply this to the cases when we partially inline a level via `tryBreakLastLevel`.

We should always respect the last dot column limit, even when we inline a level.

dansanduleac:
## After this PR
==COMMIT_MSG==
Partially inlined levels are also subject to the 'columnLimitBeforeLastBreak', that disallows the last dot in method call chains from exceeding 80 characters.

Fixes #100
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-long-method-reference.output
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-long-method-reference.output
@@ -3,8 +3,9 @@ class PalantirLongArrayInit {
         URL[] implementationUrls = implementationClassPath
                 .map(implementationUris -> {
                     log.debug("Using palantir-java-format implementation defined by URIs: {}", implementationUris);
-                    return implementationUris.stream().map(PalantirCodeStyleManager::toUrlUnchecked).toArray(URL[]
-                            ::new);
+                    return implementationUris.stream()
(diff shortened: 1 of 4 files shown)


palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-long-method-reference.output
```diff
@@ -3,8 +3,9 @@ class PalantirLongArrayInit {
         URL[] implementationUrls = implementationClassPath
                 .map(implementationUris -> {
                     log.debug("Using palantir-java-format implementation defined by URIs: {}", implementationUris);
-                    return implementationUris.stream().map(PalantirCodeStyleManager::toUrlUnchecked).toArray(URL[]
-                            ::new);
+                    return implementationUris.stream()
+                            .map(PalantirCodeStyleManager::toUrlUnchecked)
+                            .toArray(URL[]::new);
```
dansanduleac, line 8: just as a bonus, this happens to fix this particular degenerate case too

ferozco (approved):

PR #104, opened 2019-12-10
Fix string concats breaking weirdly

dansanduleac:
## Before this PR

String concatenations (or other binary operations) could be broken inside a method's multiple arguments that were otherwise inlined, and didn't look good.

See https://github.com/palantir/conjure-java-runtime/pull/1285/files#r356006884

## After this PR
==COMMIT_MSG==
No longer break string concats unless they are a method's only argument on the line so far.
==COMMIT_MSG==


ferozco (approved):

iamdanfox:
🚂

PR #107, opened 2019-12-11
Spotless plugin

dansanduleac:
## Before this PR

Logic for how to format a project with PJF is coupled between palantir-java-format and gradle-baseline.
This makes it hard for users to apply the formatter gradle plugin directly, because it needs the wiring from `com.palantir.baseline-format` to actually set up all the formatting tasks.

## After this PR
==COMMIT_MSG==
Move spotless configuration logic into a new plugin `com.palantir.java-format-spotless`. The main `com.palantir.java-format` plugin now applies both the spotless and idea plugins too.
==COMMIT_MSG==


dansanduleac:
It's sad that gradle-palantir-java-format now has to depend on the spotless plugin (not compileOnly), but this seems to be the only straightforward way to configure that plugin.

dansanduleac:
Update: was able to make it work with compileOnly - it's trickier to set this up when testing, but when using the plugin, since all the dependencies are resolved together, this will just work.

CRogers (approved):

PR #123, opened 2020-01-13
Nested method chains also adhere to column limit

dansanduleac:
## Before this PR

#70 (and #71) failed to account for more complex expressions that contained a nested chain of method calls.

An example where it didn't do the right thing is [this test](https://github.com/palantir/palantir-java-format/compare/ds/long-method-chain-arg?expand=1#diff-e506d65564d6ba3a61926e5b6ac5cabf).

## After this PR
==COMMIT_MSG==
Nested method chains now also adhere to column limit.

Also, we've made the logic a bit more specific in terms of what are method chains.
We now only apply this restricted column limit to breaks that come before actual method calls, not field accesses or parts of a fully qualified class name.
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
@@ -123,14 +123,14 @@ public final class Level extends Doc {
      *     maxWidth}
      */
     private Optional<Integer> tryToFitOnOneLine(int maxWidth, State state, Iterable<Doc> docs) {
-        float column = state.column();
-        float columnBeforeLastBreak = 0f; // Not activated by default
+        int column = state.column();
+        int columnBeforeLastBreak = 0; // Not activated by default
         for (Doc doc : docs) {
             if (doc instanceof Break && ((Break) doc).hasColumnLimit()) {
                 columnBeforeLastBreak = column;
             } else if (doc instanceof Level) {
                 // Levels might have nested levels that have a 'columnLimitBeforeLastBreak' set, so recurse.
-                State newState = state.withColumn((int) column);
+                State newState = state.withColumn(column);
                 Level innerLevel = (Level) doc;
                 Optional<Integer> newWidth = innerLevel.tryToFitOnOneLine(maxWidth, newState, innerLevel.getDocs());
                 if (!newWidth.isPresent()) {
@@ -148,9 +148,8 @@ public final class Level extends Doc {
         }
 
         // Check that the entirety of this level fits on the current line.
-        float thisWidth = getWidth(docs);
-        if (state.column() + thisWidth <= maxWidth) {
-            return Optional.of(state.column() + (int) thisWidth);
+        if (column <= maxWidth) {
+            return Optional.of(column);
         }
         return Optional.empty();
     }
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
@@ -124,9 +124,9 @@ public final class Level extends Doc {
      */
     private Optional<Integer> tryToFitOnOneLine(int maxWidth, State state, Iterable<Doc> docs) {
         float column = state.column();
-        float columnBeforeLastBreak = column;
+        float columnBeforeLastBreak = 0f; // Not activated by default
         for (Doc doc : docs) {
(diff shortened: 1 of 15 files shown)


palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-lamda-cast.output
```diff
@@ -10,7 +10,7 @@ public class Foo {
     }
 
     static String baz(Optional<String> optional) {
-        return optional.orElseGet(
-                (String) () -> bar(Optional.of("some thing that is very very very looooooong")).get());
+        return optional.orElseGet((String) () ->
+                bar(Optional.of("some thing that is very very very looooooong")).get());
```
dansanduleac, line 14: This changed beause it doesn't allow the dot in the `bar(...).get()` expression to exceed 80 chars anymore. It now ends up at exactly 80 chars, and to prove that, if I add another character to the string, then the `.get()` gets pushed onto the next line, as in:

```java
return optional.orElseGet((String) () -> bar(Optional.of("some thing that is very very very loooooooong"))
        .get());
```

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/B32114928.output
```diff
@@ -1,6 +1,7 @@
 class B32114928 {
     {
-        Class<T> tClass = (Class<T>) verifyNotNull((ParameterizedType) getClass().getGenericSuperclass())
-                .getActualTypeArguments()[0];
+        Class<T> tClass = (Class<T>)
+                verifyNotNull((ParameterizedType) getClass().getGenericSuperclass())
+                        .getActualTypeArguments()[0];
```
dansanduleac, line 5: Unclear if we want to be so strict with all methods, or if we should exempt non-side-effect-sounding methods like those that start with `get[A-Z0-9]`. @iamdanfox thoughts?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -123,25 +123,33 @@ public State computeBreaks(CommentsHelper commentsHelper, int maxWidth, State st
...
-            for (Doc doc : docs) {
-                if (doc instanceof Break) {
-                    widthBeforeLastBreak = width;
+        int column = state.column();
+        int columnBeforeLastBreak = 0; // Not activated by default
+        for (Doc doc : docs) {
+            if (doc instanceof Break && ((Break) doc).hasColumnLimit()) {
+                columnBeforeLastBreak = column;
+            } else if (doc instanceof Level) {
+                // Levels might have nested levels that have a 'columnLimitBeforeLastBreak' set, so recurse.
```
ferozco, line 132: This recursion seems like it could end up being pretty expensive

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -123,25 +123,33 @@ public State computeBreaks(CommentsHelper commentsHelper, int maxWidth, State st
...
+                State newState = state.withColumn(column);
+                Level innerLevel = (Level) doc;
+                Optional<Integer> newWidth = innerLevel.tryToFitOnOneLine(maxWidth, newState, innerLevel.getDocs());
+                if (!newWidth.isPresent()) {
+                    return Optional.empty();
```
ferozco, line 137: why do we return early here?

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
dansanduleac: the integration tests take about 15s with this block commented out, and 20s with it as it is, so I think this is fine. (it's pretty much just the extremely degenerate test case that suffers)

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
dansanduleac: It indicates that the innerLevel was not inlineable. 
Currently, the only reason for that to happen is that it had a non-prefix method call (i.e. not something like `foo.bar().stream()`) that occurs after the column limit.
Once this happens it's no point trying to one line the parent level anymore.

ferozco:
👍

PR #125, opened 2020-01-14
Keep javadoc inline tags like {@link ...} together

dansanduleac:
## Before this PR

JavadocFormatter splits inline tags with spaces, for example `{@link com.palantir.SomeClass}`, such that they can end up like
```java
/**
 * Foo bar {@link
 * com.palantir.SomeClass}
 */
```

## After this PR
==COMMIT_MSG==
In javadocs, keep inline tags like `{@link ...}` on one line if not already split.
==COMMIT_MSG==


palantir-java-format/src/main/java/com/palantir/javaformat/java/javadoc/JavadocLexer.java
```diff
@@ -160,13 +162,13 @@ private Type consumeToken() throws LexException {
...
             braceDepth.increment();
-            return LITERAL;
+            return INLINE_TAG_OPEN;
         } else if (input.tryConsume("{")) {
             braceDepth.incrementIfPositive();
             return LITERAL;
         } else if (input.tryConsume("}")) {
             braceDepth.decrementIfPositive();
-            return LITERAL;
+            return braceDepth.isPositive() ? LITERAL : INLINE_TAG_CLOSE;
```
CRogers, line 171: What happens here with nested inline tags? I noticed there's an `inlineTagDepth` in the other file, but I would have thought that this would only make these tokens for the outermost (from my brief can of these changes)?

CRogers: Do we even care about nested inline tags?

robert3005 (approved):

palantir-java-format/src/main/java/com/palantir/javaformat/java/javadoc/JavadocLexer.java
dansanduleac: we care for situations like
```
{@code
class Foo {
}
}<pre>
```

as shown in one of the tests I added, that's actually copied from one of the integration tests where this was initially being handled incorrectly.
Specifically, we want to treat the closing brace of the *class* as a literal (i.e. this is still literal stuff inside the code block), and only the last brace as a INLINE_TAG_CLOSE.

dansanduleac: On the other hand, in the other method, I use another one of these depth counters but I don't *think* the number can be anything other than 0 or 1.

CRogers: Makes sense

CRogers:
:+1:

PR #138, opened 2020-01-21
Fix column limit logic to work as expected with initializers

dansanduleac:
## Before this PR

In a Level representing an assignment initializer (such as a variable or field declaration), there is an expectation that the initializer expression follows an optional break after the `=` or `:` (in the case of a for-each) sign.

We try to break that level first to see if just breaking once allows the rest of the initializer to fit on one line, and if so, we prefer that, instead of partially inlining the expression onto the `=` line, which may end up spanning more than 2 lines.

However, some shortcut logic for "does this chunk of code fit on the current line" was flawed in that it didn't take into account potential breaks due to column limits in method chains (which became a problem after #123). Therefore, at one point it decided it fits so we don't need to break after the `=`, but later down the line, it would end up breaking the code anyway because of the column limit.

## After this PR
==COMMIT_MSG==
Method chains in initializers may once again be laid onto a single next line, if they are short enough. This fixes a regression from #123 (0.3.11) where the column limit for method chains would prevent us from attempting to put the initializer on the 2nd line first and see if that works.
==COMMIT_MSG==

## Possible downsides?

Fewer builders are forcibly split onto multiple lines.
However, if we want to be more aggressive with pushing this behaviour for builders, we should do that differently, not as a side-effect of this bug.

### DIFF
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/B20701054.output
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/B20701054.output
@@ -14,29 +14,16 @@ class B20701054 {
         ImmutableList<String> x = ImmutableList.INSTANCE.add(1).build();
         ImmutableList<String> x = ImmutableList.INSTANCE.add(1).add(2).build();
         ImmutableList<String> x = ImmutableList.INSTANCE.add(1).add(2).add(3).build();
-        ImmutableList<String> x = ImmutableList.INSTANCE
-                .add(1)
-                .add(2)
-                .add(3)
-                .add(4)
-                .build();
+        ImmutableList<String> x =
+                ImmutableList.INSTANCE.add(1).add(2).add(3).add(4).build();
 
         ImmutableList<String> x = ImmutableList.builder().add(1).build();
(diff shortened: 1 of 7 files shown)


palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/B20701054.output
```diff
@@ -14,29 +14,16 @@ class B20701054 {
...
-                .add(1)
-                .add(2)
-                .add(3)
-                .add(4)
-                .add(5)
-                .build();
+        ImmutableList<String> x =
+                ImmutableList.builder().add(1).add(2).add(3).add(4).build();
+        ImmutableList<String> x =
+                ImmutableList.builder().add(1).add(2).add(3).add(4).add(5).build();
```
CRogers, line 26: Are these examples really better? It's harder to scan chained calls horizontally than when they're lined up vertically.

CRogers: (apparently this keeps it consistent with other parts of the formatter, rather than letting them fight)

CRogers (approved):

PR #144, opened 2020-01-21
wait longer for git diff

dansanduleac:
## Before this PR

`git diff` on large repos fails to return in 10 seconds.

## After this PR
==COMMIT_MSG==
Wait 30 seconds for git diff output before giving up.
==COMMIT_MSG==


iamdanfox (approved):
Oof. Where did you observe this?

iamdanfox:
Edit nvm just saw the internal data ingestion product.

PR #147, opened 2020-01-23
Don't always inline expression lambdas

dansanduleac:
## Before this PR

We always allow an inlining to stop at a lambda, regardless of the complexity of the expression following the lambda.

This can sometimes lead to inlining a bunch of arguments but breaking up a super simple lambda, which is not readable at all:

```java
publisher = new ThisThatEventPublisher(
        txnManager, () -> Events.encoder(keyPair, new TestClock()), EventSeries.THING, () ->
                DEFAULT_EVENT_BATCH_SIZE);
```

## After this PR
==COMMIT_MSG==
More strictly control when we can inline a chunk of code followed by an expression lambda.

Specifically, the inlining chain now continues through to the body of the expression lambda, with the exception that the inlined expression so far is reset to be considered "simple" when entering a lambda expression.

The exception is necessary in order to preserve the inlining of [palantir-chains-lambdas](https://github.com/palantir/palantir-java-format/compare/ds/better-lambdas?expand=1#diff-08ee473a0a14f711ab84d86b527a9725R3) which we consider desirable.
==COMMIT_MSG==

Example changes on:
- [conjure-java-runtime](https://github.com/palantir/conjure-java-runtime/pull/1449/files/2011fcd216df2c7cb2e95ef0cbaed9961e3b320c..277f401b862086b7f62aace6c29b62e736517fe1)

## Possible downsides?
* Inadvertently changing more code that looked better before.
I think this is acceptable as long as the existing code doesn't become _less_ readable but just a bit more sparse, if in return we get rid of some corner cases that were extremely unreadable / confusing.


dansanduleac:
Holding off on this for a bit until we can determine how many changes it will cause in a large internal repo.

iamdanfox (approved):

PR #152, opened 2020-01-27
Turn off the JavadocComment

dansanduleac:
## Before this PR

The javadoc formatter is too opinionated for a lot of our repos and causes disruption while people manually clean up legacy docstrings that have markdown-style lists or carefully laid out code that should live inside a `<pre>` but doesn't.

## After this PR
==COMMIT_MSG==
Turn off the javadoc formatter entirely.

We still indent the javadoc prefix to the corect column, but don't do any code reflowing or HTML tag correction anymore.
==COMMIT_MSG==


iamdanfox (approved):

iamdanfox:
Just some more tests to go green?

PR #158, opened 2020-01-28
service load only once

dansanduleac:
## Before this PR

spotless is service loading the implementation once per gradle project, even though it's guaranteed to be the same for the entire gradle build.

This leads to some people's gradles dying by exceeding MaxMetaspaceSize.

## After this PR
==COMMIT_MSG==
Only service load the formatter implementation once.
==COMMIT_MSG==

### DIFF
--- a/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/JavaFormatExtension.java
+++ b/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/JavaFormatExtension.java
@@ -16,22 +16,30 @@
 
 package com.palantir.javaformat.gradle;
 
+import com.google.common.base.Suppliers;
 import com.google.common.collect.Iterables;
 import com.palantir.javaformat.java.FormatterService;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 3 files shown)


gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/JavaFormatExtension.java
```diff
@@ -16,22 +16,30 @@
...
 import com.google.common.collect.Iterables;
 import com.palantir.javaformat.java.FormatterService;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ServiceLoader;
+import java.util.function.Supplier;
 import org.gradle.api.artifacts.Configuration;
 
 public class JavaFormatExtension {
```
iamdanfox, line 29: :+1: confirmed this is only registered by the PalantirJavaFormatProvider plugin, which only works on the root project:

```
        Preconditions.checkState(
                rootProject == rootProject.getRootProject(),
                "May only apply com.palantir.java-format-provider to the root project");
```

iamdanfox (approved):

PR #169, opened 2020-02-04
Better lambda/assignment inlining logic

dansanduleac:
## Before this PR

#147 caused a slight regression where we'd now eagerly inline some prefix of a complex expression if it's a lambda expression body or an initializer (e.g. assignment), if the expression in question doesn't fit _entirely_ on the same line or on the 2nd line.

Example: https://github.com/palantir/conjure-java/pull/721/files#diff-407b18e644dee71e5e305db5d4d2be0fR324

This is not ideal as it makes it harder to read.

## After this PR
==COMMIT_MSG==
Reduce situations where an inlined lambda can get a newline after the arrow, and allow inlining new Class() calls in more contexts.

This is a two pronged change:
* **When part of a simple inlining**: prefer partially inlining (instead of putting entire lambda body expression on the 2nd line) only if this results in fewer lines used, *or* it's a long method chain, like a builder.
* **When part of a complex inlining**: partially inline iff it fits and the expression being inlined is complex, otherwise abort the inlining chain

-- simple / complex refer to the complexity of the levels partially inlined so far in the given inlining chain, where if any of the levels is complex, the inlining is considered complex.
==COMMIT_MSG==

Example change:
- https://github.com/palantir/conjure-java/pull/730


dansanduleac:
Ran it on conjure-java and a large internal repo and every change seems net positive.

palantir-java-format/src/main/java/com/palantir/javaformat/doc/Level.java
```diff
@@ -296,8 +296,14 @@ private Exploration breakNormally(State state, LevelNode levelNode, CommentsHelp
         if (keepIndent) {
             newState = newState.withIndentIncrementedBy(getPlusIndent());
         }
-        return Optional.of(tryToLayOutLevelOnOneLine(
-                commentsHelper, maxWidth, newState, memoizedSplitsBreaks.get(), explorationNode));
+        State partiallyInlinedState = tryToLayOutLevelOnOneLine(
```
iamdanfox: ~why not move this down under the bodyIsComplex calculation so we can short-circuit and avoid computing it if bodyIsComplex == false?~

EDIT nvm misread the if

iamdanfox:
Output looks nice - could you merge https://github.com/palantir/palantir-java-format/pull/172 and then update from develop so we capture the effects?

iamdanfox (approved):

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -2625,8 +2625,10 @@ void visitDot(ExpressionTree node0) {
                         .debugName("visitDot")
                         .plusIndent(plusFour)
                         .breakBehaviour(BreakBehaviours.preferBreakingLastInnerLevel(true))
-                        .breakabilityIfLastLevel(LastLevelBreakability.ACCEPT_INLINE_CHAIN)
+                        .breakabilityIfLastLevel(
+                                LastLevelBreakability.ACCEPT_INLINE_CHAIN_IF_SIMPLE_OTHERWISE_CHECK_INNER)
```
dansanduleac, line 2632: this fixes palantir-5 and B18479811

dansanduleac:
👍

PR #172, opened 2020-02-05
A few more tests

iamdanfox:
I wanted a few more examples to be able to preview the effects of https://github.com/palantir/palantir-java-format/pull/169, just putting these in a separate PR so that we can more easily preview the _changes_ that PR causes.


dansanduleac:
👍

PR #191, opened 2020-02-13
inline annotation args

dansanduleac:
## Before this PR

Long annotation args would be split onto the next line.

## After this PR
==COMMIT_MSG==
Long annotation args are now partially inlined, such that concatenated strings look better.
==COMMIT_MSG==


iamdanfox (approved):
Oh this is nice!

PR #192, opened 2020-02-13
Respect user's existing splits in string concatenations

dansanduleac:
## Before this PR

We'd split long string concatenations such that we'd end up with one item per line.
This breaks carefully crafted Sql statements in annotations or other places.

## After this PR
==COMMIT_MSG==
Preserve the user's newlines in a long string concatenation (defined as a concatenation with `+` where at least one expression is a literal string).

Also allow inlining annotation values.
==COMMIT_MSG==

## Possible downsides?

cc @sixinli


palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -3148,12 +3157,35 @@ private boolean isFormatMethod(List<? extends ExpressionTree> arguments) {
...
             return false;
         }
-        return isStringConcat(arguments.get(0));
+        return isFormatString(arguments.get(0));
     }
 
     private static final Pattern FORMAT_SPECIFIER = Pattern.compile("%|\\{[0-9]\\}");
 
     private boolean isStringConcat(ExpressionTree first) {
+        final boolean[] stringConcat = {false};
```
iamdanfox, line 3168: ah yes, the poor man's AtomicBoolean 😂

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -1144,18 +1144,24 @@ public Void visitBinary(BinaryTree node, Void unused) {
...
+        FillMode nextFillMode = builder.lastTokenFollowedByNewline() ? UNIFIED : fillMode;
         int operatorsN = operators.size();
         for (int i = 0; i < operatorsN; i++) {
-            builder.breakOp(fillMode, " ", ZERO);
+            builder.breakOp(nextFillMode, " ", ZERO);
             builder.op(operators.get(i));
+            boolean shouldEnforceNewline = builder.lastTokenFollowedByNewline();
             builder.space();
             scan(operands.get(i + 1), null);
+            shouldEnforceNewline |= builder.lastTokenFollowedByNewline();
```
iamdanfox: it seems like `|=` is a non-short circuiting operator... is that intentional? https://stackoverflow.com/a/2486574/1941552

palantir-java-format/src/main/java/com/palantir/javaformat/OpsBuilder.java
```diff
@@ -303,6 +306,12 @@ public void close() {
         return peekToken(0);
     }
 
+    /** Return whether the last token emitted is followed by a newline. */
+    public boolean lastTokenFollowedByNewline() {
```
iamdanfox: s/lastToken/mostRecentToken/ ? I found the `last` thing a tiny bit ambiguous (like is this referring to the very last one in a long list somehow)

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-10.output
```diff
@@ -0,0 +1,10 @@
+class Palantir10 {
+    @SqlUpdate("CREATE TABLE things "
+            + "(id VARCHAR(" + MAX_LENGTH + ") NOT NULL, "
+            + "title VARCHAR(" + MAX_LENGTH + ")  NOT NULL, "
+            + "description VARCHAR(" + MAX_DESCRIPTION_LENGTH + ") NOT NULL, "
+            + "standard BOOLEAN NOT NULL, "
+            + "documentJson VARCHAR(" + MAX_DOCUMENT_LENGTH + ") NOT NULL, "
+            + "PRIMARY KEY (id))")
```
iamdanfox, line 9: I wonder if we're going to lose some otherwise nice reformatting by optimizing for this case where there are 14 `+` symbols... like what about short snippets with just a little concatenation?

```java
doSomething("foo" +
 bar + "baz")
```

Maybe we should count the +s?

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
ferozco: Perhaps a little comment here explaining we expect the builder to be mutated between calls

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-10.output
ferozco: The formatter seems to behave well in small cases as well. It allowed
```
class Palantir10 {
    @SqlUpdate("CREATE TABLE things " + MAX_LENGTH + ")")
    void createTable();
}
```

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
dansanduleac: I guess I didn't mind, but sure, can save a bit of work.

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/palantir-10.output
dansanduleac: ok, how about this - one newline => don't bother, more => keep

dansanduleac: added tests to showcase @iamdanfox's  concern is now handled well

dansanduleac: @ferozco the small case you pasted worked before as well, since the annotation fits onto one line there is no need to ever split it.
This is only tricky when the whole annotation can't fit on a single line.

iamdanfox (approved):

dansanduleac:
🐞

PR #193, opened 2020-02-13
Fix lineSpan

dansanduleac:
## Before this PR

#192 added a function `lineSpan` that could sometimes throw a NPE.

## After this PR
==COMMIT_MSG==
Fix lineSpan function to not throw NPE.
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
@@ -85,7 +85,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
@@ -3307,8 +3306,6 @@ public final class JavaInputAstVisitor extends TreePathScanner<Void, Void> {
     int lineSpan(Tree node) {
         ImmutableRangeMap<Integer, ? extends Input.Token> positionTokenMap =
                 builder.getInput().getPositionTokenMap();
-        Function<Input.Token, Integer> lineNumberAt =
-                token -> builder.getInput().getLineNumber(token.getTok().getPosition());
 
         int startPosition = getStartPosition(node);
         int endPosition = getEndPosition(node, getCurrentPath());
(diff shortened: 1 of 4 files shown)


palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -3304,13 +3305,22 @@ private Integer actualColumn(ExpressionTree expression) {
...
     /** How many lines does this node take up in the input. Returns at least 1. */
     int lineSpan(Tree node) {
-        IntFunction<Integer> lineNumberAt = tokenPosition -> {
-            Input.Token token = builder.getInput().getPositionTokenMap().get(tokenPosition);
-            return builder.getInput().getLineNumber(token.getTok().getPosition());
-        };
-        return lineNumberAt.apply(getEndPosition(node, getCurrentPath()))
-                - lineNumberAt.apply(getStartPosition(node))
-                + 1;
+        ImmutableRangeMap<Integer, ? extends Input.Token> positionTokenMap =
```
ferozco, line 3307:
```suggestion
        RangeMap<Integer, ? extends Input.Token> positionTokenMap =
```

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -3304,13 +3305,22 @@ private Integer actualColumn(ExpressionTree expression) {
...
+                builder.getInput().getPositionTokenMap();
+        Function<Input.Token, Integer> lineNumberAt =
```
ferozco: can we make this an instance method?

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
dansanduleac: I thought immutables recommend using them as types directly

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
dansanduleac: sure

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
iamdanfox: Yeah I think there is actually an error-prone rule for this, because using the `Immutbale*` type ensures that if anyone further down tries to call one of the mutator methods (e.g. put) then they get a nice deprecation message.

iamdanfox (approved):

PR #203, opened 2020-02-17
Fix crash when breaking up long comment after simple lambda body

dansanduleac:
## Before this PR

A long comment that follows an otherwise simple lambda body (no inner levels) can cause a crash if it ends up being chunked onto multiple lines by the `CommentsHelper`.

```java
arg -> true // long comment might end up
            // being chunked oops
```

## After this PR
==COMMIT_MSG==
Make lambda/assignment logic more resilient so it doesn't crash when encountering long comments.
==COMMIT_MSG==

## Possible downsides?

This currently means that if the lambda / assignment body is not complex but has a long comment which doesn't fit on one line, the body will always end up on the 2nd line:
```java
arg ->
        true // long comment might end up
             // being chunked oops
```

If this is not desirable, we could revert the heuristic to what it was before, but be resilient in the next method down the line `tryInlinePrefixOntoCurrentLine` so we don't crash if no inner level is found.

### DIFF
--- a/changelog/@unreleased/pr-203.v2.yml
+++ b/changelog/@unreleased/pr-203.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Make lambda/assignment logic more resilient so it doesn't crash when
+    encountering long comments.
+  links:
+  - https://github.com/palantir/palantir-java-format/pull/203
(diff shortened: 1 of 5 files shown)


ferozco:
Instead of breaking the line because of trailing comment could we move the comment to be on a separate line above the simple lambda?

dansanduleac:
We possibly could, but that's a pretty big deviation from how the formatter was set up to deal with comments, so it might break other things

ferozco:
ok, if you don't think its worth it 👍

PR #252, opened 2020-03-27
Inline case statements with single block body

dansanduleac:
## Before this PR

Case statements with a block body get unnecessarily put onto the next line, causing checkstyle to complain.

Sometimes it's not possible to avoid this pattern, if you are defining variables inside the case statement.

## After this PR
==COMMIT_MSG==
Case statements with a body consisting of a single block now inline their block.
==COMMIT_MSG==


iamdanfox (approved):

dansanduleac:
pong

PR #257, opened 2020-03-30
Fix possible exception with initializer body close to column limit

dansanduleac:
## Before this PR

When formatting far-off (close to 120 characters) lambda or assignment body, the formatting code could throw an exception.
This would happen if the body contained an empty level, such as the case when e.g. doing `new SomeClass<>()`: there is an empty level created for the type arguments inside `<>`, but it's empty because there are no type arguments. 

This would throw off the logic that decides whether trying to put such body onto the next line results in it taking only one line (in `handle_breakOnlyIfInnerLevelsThenFitOnOneLine`).
It could first infer that the empty level was broken (marked as such because the column would have already exceeded 120 columns), but then later down the line `tryInlinePrefixOntoCurrentLine` would try to find this broken level, only this time it would filter out empty levels, and so never find that level and throw.

## After this PR
==COMMIT_MSG==
Fix edge case with empty levels causing formattings very close to the column limit to throw.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-257.v2.yml
+++ b/changelog/@unreleased/pr-257.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fix edge case with empty levels causing formattings very close to the
+    column limit to throw.
+  links:
+  - https://github.com/palantir/palantir-java-format/pull/257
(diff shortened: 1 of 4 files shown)


carterkozak (approved):
Thanks, Dan!

PR #351, opened 2020-09-21
Add runtime dependency on palantir-java-format

jakobjuelich:
## Before this PR
The build.gradle did not specify a dependency on com.palantir.javaformat:palantir-java-format, but PalantirJavaFormatProviderPlugin.java does add it during runtime.

## After this PR

==COMMIT_MSG==
Specify gradle-palantir-java-format's dependency on palantir-java-format
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-351.v2.yml
+++ b/changelog/@unreleased/pr-351.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Specify gradle-palantir-java-format's dependency on palantir-java-format
+  links:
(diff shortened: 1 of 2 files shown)


jakobjuelich:
Won't work because of https://github.com/palantir/palantir-java-format/blob/18dc33929edba0959025ea8e69fc42f013ec80f7/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L83 :/

PR #365, opened 2020-11-16
Add support for java 14 language features

### LINKED ISSUE #347: Support Java 14 language features (and decouple from javac-shaded)
## What happened?

Things like records, switch expressions, a new `instanceof` syntax and multiline strings are available in newer java compilers.

Support for these features landed in upstream google-java-format in various commits, including https://github.com/google/google-java-format/commit/4ddb9148c103b9c65768e519884b0c8ad5caaf39.

Seems like they did this by making a new `Java14InputAstVisitor` which extends the normal `JavaInputAstVisitor` and then reflectively deciding whether or not to construct it depending on what java compiler version is on the classpath, as this affects whether `org.openjdk.source.util.TreePathScanner` has the extra visit methods or not:

- visitBindingPattern
- visitYield
- visitSwitchExpression
- visitInstanceOf

I _think_ we currently bundle `com.google.errorprone:javac-shaded:9+181-r4173-1` in p-j-f, which means we're going to be a bit stuck on this front.  Google switched away from this in https://github.com/google/google-java-format/pull/447, and I think we need to do the same in order to be able to keep picking up new language features.

This will likely also require some tweaks to how p-j-f is _used_ from the intellij plugin etc, as we'll need to ensure the relevant java _modules_ (com.sun.tools.javac.{file,main,parser,tree,util,code,api}) are exposed using something like the [`--add-exports` flag](https://github.com/google/google-java-format/pull/447/files#diff-357e4854869b2e21c38b1b437f11095aR94).

cc @fawind

fawind:
## After this PR

Closes https://github.com/palantir/palantir-java-format/issues/347

Cherry-picking relevant commits from [google-java-format](https://github.com/google/google-java-format/compare/google-java-format-1.7...google-java-format-1.9) and adapting them for pjf.

Changes:
* Use build in jdk javac instead of errorprone javac.
  * Because of this, the source and target compatibility is increased to 11.
  * That's also the reason for all the import changes.
* Add initial support for switch-statements, var, yield, records, text blocks.

==COMMIT_MSG==
Add support for java 14 language features
==COMMIT_MSG==

## Possible downsides?
Dropping jdk 8 support.


debugger/package-lock.json
```diff
@@ -0,0 +1,13599 @@
+{
```
iamdanfox: I think you can zap this file as it contains some internal links, and yarn.lock should be sufficient

palantir-java-format/build.gradle
```diff
@@ -32,6 +31,30 @@ dependencies {
...
+        '--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED',
+        '--add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED',
+        '--add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED',
+        '--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED',
+        '--add-exports=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED',
+        '--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED'
+    ]
+
+    if (JavaVersion.current() < JavaVersion.VERSION_14) {
+        excludes = ['**/Java14InputAstVisitor.java']
```
iamdanfox, line 47: oooh nice, i was wondering how painful it was gonna be to do some kinda of dual compilation here, or if were going to have to do some multi-release jar thing!

One caveat with this approach is I think gradle has some new functionality where you can specify the JDK to use for these compile tasks independently from the JDK used to _evaluate_ gradle, and JavaVersion.current() will grab the gradle evaluating one

palantir-java-format/src/test/java/com/palantir/javaformat/java/FileBasedTests.java
```diff
@@ -71,6 +75,9 @@ public FileBasedTests(Class<?> testClass, String testDirName) {
                         .isEqualTo(1);
                 String baseName = getNameWithoutExtension(subPath.getFileName().toString());
                 String extension = getFileExtension(subPath.getFileName().toString());
+                if (JAVA_14_TESTS.contains(baseName) && Formatter.getMajor() < 14) {
+                    continue;
+                }
```
iamdanfox: I'd recommend junit's assumeThat - it can give you a really nice clear 'skipped' thing in the intellij ui

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.output
```diff
@@ -0,0 +1,21 @@
...
+            }
+        };
+    }
+
+    int yield() {
+        return switch (i) {
+            case 0:
+                yield 0;
+            default:
+                yield n / i;
```
iamdanfox, line 18: woah this is a funky keyword... ~based on the reading of https://www.baeldung.com/java-thread-yield it sounds like not something that we expect palantir codebases to contain!~

nvm Thread.yield and `yield` keyword are _totally_ different... looks like this one is just a siwtch-expression specific return!

palantir-java-format/src/main/java/com/palantir/javaformat/java/Formatter.java
```diff
@@ -197,6 +211,24 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
...
+            return (int) version.getClass().getMethod("major").invoke(version);
+        } catch (Exception e) {
+            // continue below
+        }
+        int version = (int) Double.parseDouble(JAVA_CLASS_VERSION.value());
+        if (49 <= version && version <= 52) {
+            return version - (49 - 5);
+        }
+        throw new IllegalStateException("Unknown Java version: " + JAVA_SPECIFICATION_VERSION.value());
+    }
```
iamdanfox, line 214: this looked a bit bonkers to me, but I see it matches what they have in g-j-f!

https://github.com/google/google-java-format/blob/4ddb9148c103b9c65768e519884b0c8ad5caaf39/core/src/main/java/com/google/googlejavaformat/java/Formatter.java#L179-L194

build.gradle
```diff
@@ -45,8 +46,8 @@ subprojects {
     apply plugin: 'java-library'
     apply plugin: 'org.inferred.processors'
 
-    sourceCompatibility = 1.8
-    targetCompatibility = 1.8
+    sourceCompatibility = 11
+    targetCompatibility = 11
```
iamdanfox, line 50: this probably warrants a callout in the changelog, that palantir-java-format is now only able to run on java11+

debugger/package-lock.json
fawind: removed

build.gradle
fawind: For sure! Might even publish a new major release so 2.x is for jdk11+

palantir-java-format/src/main/java/com/palantir/javaformat/java/java14/Java14InputAstVisitor.java
```diff
@@ -0,0 +1,237 @@
...
+import com.sun.tools.javac.code.Flags;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.tree.TreeInfo;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Extends {@link JavaInputAstVisitor} with support for AST nodes that were added or modified for Java 14.
+ */
+public class Java14InputAstVisitor extends JavaInputAstVisitor {
```
iamdanfox, line 45: I'm wondering what the longer-term implications of this pattern are - e.g. java 15 introduces a teeny bit more syntax right (specifically the `sealed` keyword I think)... would the idea be to have one of these classes for each of the java versions we support compiling with? and then at some point take a hard stance and increment our required min version, collapsing some of these classes into one?

.circleci/template.sh
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
 export CIRCLECI_TEMPLATE=java-library-oss
-export PRIMARY_BRANCH=develop
+export ONLY_11=true
+export UNIT_TEST_15=true
+export UNIT_TEST_14=false
```
iamdanfox: I think for the final publish step, it's important the the java compiler is java 14 to ensure we don't hit the exclusion below, as we definitely want that class to appear in the final jar!

```gradle
    if (JavaVersion.current() < JavaVersion.VERSION_14) {
        excludes = ['**/Java14InputAstVisitor.java']
    }
```

Maybe we have to diverge from the CircleCI template?

Or maybe JDK=14 / JDK=15 would be enough? Seems like we do want. the entirely independent full `./gradlew build` pipelines on both java11, java14 and maybe other higher versions too? (i.e. effectively our unit-test-11/unit-test-15 thingies I think).

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -3396,7 +3412,10 @@ int declareOne(
             if (modifiers.isPresent()) {
                 visitAndBreakModifiers(modifiers.get(), annotationsDirection, Optional.of(verticalAnnotationBreak));
             }
-            builder.open(type != null ? plusFour : ZERO);
+            boolean isVar = builder.peekToken().get().equals("var")
```
iamdanfox, line 3415: ok so this is the first block where we diverge from the google-java-format version? https://github.com/google/google-java-format/commit/4ddb9148c103b9c65768e519884b0c8ad5caaf39#diff-0c133fd389bc6f207e146ba08dbc50dd78aa088ff31cdc56cb14ebb31e97ec49R158

EDIT nvm seems like we're up to date with their current impl, not their initial PR

https://github.com/google/google-java-format/blob/93d649d0fbc07fc604f7b47b3098cf6e1df521de/core/src/main/java/com/google/googlejavaformat/java/JavaInputAstVisitor.java#L3287-L3291.

It's kinda interesting because they're using the peekToken and doing a string comparison to get away with supporting java13+ features here in this plain JavaInputAstVisitor, rather than making another verbose Java13InputAstVisitor.

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -3444,10 +3465,15 @@ int declareOne(
                     initializer.get().accept(this, null);
                     builder.close();
                 } else {
-                    builder.open(
-                            Indent.If.make(typeBreak, plusFour, ZERO),
-                            BreakBehaviours.breakOnlyIfInnerLevelsThenFitOnOneLine(true),
-                            LastLevelBreakability.ABORT);
+                    if (builder.peekToken().get().equals("switch")) {
+                        // TODO(fawind): Don't break switch expression assignment
+                        builder.open(Indent.If.make(typeBreak, plusFour, ZERO));
```
iamdanfox, line 3470: I tried deleting this block (leaving it as it was before) and it seemed like all the `FormatterIntegrationTest` cases passed.  Is it really necessary? if so can we add the integration test to capture it?

palantir-java-format/src/main/java/com/palantir/javaformat/java/java14/Java14InputAstVisitor.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @Override
+    public Void visitYield(YieldTree node, Void aVoid) {
+        sync(node);
+        token("yield");
+        builder.space();
+        scan(node.getValue(), null);
+        token(";");
+        return null;
+    }
```
iamdanfox, line 68: amusing that they missed the `yield` keyword in their original g-j-f PR: https://github.com/google/google-java-format/commit/4ddb9148c103b9c65768e519884b0c8ad5caaf39#diff-8dbec381ea929cc5907d0ae86ac47ce873aa4ed4ad9337ce85db28642b1dd67fR59-R63. Confirming this now matches the current state! https://github.com/google/google-java-format/blob/93d649d0fbc07fc604f7b47b3098cf6e1df521de/core/src/main/java/com/google/googlejavaformat/java/java14/Java14InputAstVisitor.java#L60-L68

palantir-java-format/src/main/java/com/palantir/javaformat/java/java14/Java14InputAstVisitor.java
```diff
@@ -0,0 +1,237 @@
...
+            case RECORD:
+                visitRecordDeclaration(tree);
+                break;
+            default:
+                throw new AssertionError(tree.getKind());
+        }
+        return null;
+    }
+
+    public void visitRecordDeclaration(ClassTree node) {
```
iamdanfox, line 99: this is pretty hard to read, but it looks like at least it matches the current g-j-f impl, so at least we'll be able to stay in sync. https://github.com/google/google-java-format/blob/93d649d0fbc07fc604f7b47b3098cf6e1df521de/core/src/main/java/com/google/googlejavaformat/java/java14/Java14InputAstVisitor.java#L99-L160

palantir-java-format/src/main/java/com/palantir/javaformat/java/java14/Java14InputAstVisitor.java
```diff
@@ -0,0 +1,237 @@
...
+                if (!first) {
+                    token(",");
+                    builder.space();
+                }
+                scan(expression, null);
+                first = false;
+            }
+        }
+        switch (node.getCaseKind()) {
+            case STATEMENT:
```
iamdanfox, line 213: looks like this branch diverges slightly - what was the motivation here?

https://github.com/google/google-java-format/blob/93d649d0fbc07fc604f7b47b3098cf6e1df521de/core/src/main/java/com/google/googlejavaformat/java/java14/Java14InputAstVisitor.java#L218-L223

palantir-java-format/src/main/java/com/palantir/javaformat/java/Formatter.java
iamdanfox: Actually hang on - we've specified our sourceCompat and targetCompat=11 now right? so we should be able to throw away all these reflective invocations, and just call Runtime.version() because that's available in java9 upwards???

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
iamdanfox: I put a debugger on this, and it _seems_ like the call stack of visitEnhancedForLoop -> visitToDeclare -> declareOne, we lost the `node` variable which could give us a first-class, typed way of detecting this rather than just relying on string comparisons.

I believe it would just be a matter of calling the `JCVariableDecl#isImplicitlyTyped()` method, and that would tell us we're looking at a `var` declaration. Not 100% sure if it's worth diverging from their implementation in order to plumb this down though, as it might make future maintenance more annoying, despite slightly cleaner code

iamdanfox:
Thanks for burning through all the thorny gradle bits to make this actually work in our setup, I'm pretty psyched to actually use these new preview features :)

I think there's the publishing thing to fix (needs a jdk14 in the docker container to compile with), then I think you should probably cut a release candidate here, and just see if there are any additional changes needed in baseline at all for this to be useable. Probably worth making your own version of that devx/template-witchcraft#1414 PR, and seeing how many hacks are strictly necessary!

palantir-java-format/src/test/java/com/palantir/javaformat/java/FileBasedTests.java
fawind: Good tip, updated!

palantir-java-format/src/main/java/com/palantir/javaformat/java/Formatter.java
fawind: Yep, this is no longer necessary. Replaced with `Runtime.version().feature()` which got introduced in jdk9

palantir-java-format/src/main/java/com/palantir/javaformat/java/java14/Java14InputAstVisitor.java
fawind: The pjf `visitStatement` method as a different signature to support inlining statements:
* pfj: https://github.com/palantir/palantir-java-format/blob/8552645373eed19de59b40262c39249ce7d91d4c/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java#L2173
* gjf: https://github.com/google/google-java-format/blob/6bde01e0880586518efad4cc5fd5a00916f41bb5/core/src/main/java/com/google/googlejavaformat/java/JavaInputAstVisitor.java#L2148

The code used here is consistent to how we handle case statements in the non-j14 visitor: https://github.com/palantir/palantir-java-format/blob/8552645373eed19de59b40262c39249ce7d91d4c/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java#L1832-L1838

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
fawind: Added a test case for that.

When using a switch-expression as an assignment and this breaking behavior:
```
builder.open(
    Indent.If.make(typeBreak, plusFour, ZERO),
    BreakBehaviours.breakOnlyIfInnerLevelsThenFitOnOneLine(true),
    LastLevelBreakability.ABORT);
```

the formatter tries to inline the last line of the statement, e.g.:

```java
int assignment(int j) {
        boolean val = switch (j) {
            case 0 -> true;
            default -> false;};    // <---
        return val;
}
```

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
fawind: Happy to make the change but I would probably stick to the original impl if possible. That way we can cherry-pick new features and fixes instead of reimplementing them which will likely save us quite a bit of work down the road.

.circleci/config.yml
```diff
@@ -119,36 +120,44 @@ jobs:
...
+          when: always
       - save_cache:
           key: 'trial-publish-gradle-cache-v2-{{ checksum "versions.props" }}-{{ checksum "build.gradle" }}'
           paths: [ ~/.gradle/caches ]
       - store_test_results: { path: ~/junit }
       - store_artifacts: { path: ~/artifacts }
 
   publish:
-    docker: [{ image: 'circleci/openjdk:8u222-stretch-node' }]
+    docker: [{ image: 'circleci/openjdk:15-jdk-buster-node' }]
```
iamdanfox, line 145: nice, so we're publishing with java 15, but unit-test-11 should ensure that older codepaths still work :)

iamdanfox (approved):
Looks great to me! Might be worth fleshing out the implications of the sourceCompat 11 thing a little bit more in the changelog though, kinda to explain that palantir-java-format 2.X can only _run_ on java11+ now (which is LTS). It's OK run the formatter on java11, but still produce java8 jars from your repo, but if you want to run the formatter on java8 you won't be able to upgrade past 1.X

fawind:
👍

PR #562, opened 2021-10-27
Bootstrap new jvm for running formatter with flags

fawind:
## Before this PR

There are two issue with running the formatter targeting newer Java versions:
1. Intellij runs plugins using its custom JDK 11 which results in the intellij plugin not being able to format files containing newer language features.
2. Starting with Java 16, the JVM running the formatter has to be run with additional `--add-export` flags due to JEP 396 ([ref](https://github.com/google/google-java-format/blob/96a8b90cc9a0bef28401823d3ae55305ea9cd6e4/README.md#jdk-16)).

## After this PR

As a workaround to both problems we bootstrap a new JVM in which we run the formatter. This allows us to specify the additional jvm flags without having to set them on the main JVM process and to run the formatter with a newer JDK than the one provided by Intellij

**FLUPS:**
* ~~Improve performance for spotless-gradle integration~~ - For now we just updated the Intellij plugin.
* Enable bootstrapping intellij formatter for projects > Java 11

==COMMIT_MSG==
Bootstrap new jvm for running formatter with flags
==COMMIT_MSG==


palantir-java-format-jre-bootstrap/src/main/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterService.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private static final String FORMATTER_MAIN_CLASS = "com.palantir.javaformat.java.Main";
+
+    private final Supplier<Path> jrePath;
+    private final Supplier<Integer> jreMajorVersion;
+    private final Supplier<List<URL>> implementationClassPath;
+
+    public BootstrappingFormatterService(
+            Supplier<Path> jrePath, Supplier<Integer> jreMajorVersion, Supplier<List<URL>> implementationClassPath) {
+        this.jrePath = jrePath;
```
carterkozak: naming nit: I believe this _must_ be a jdk, otherwise the compiler classes we need aren't present.

palantir-java-format-jre-bootstrap/src/test/java/com/palantir/java/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
```diff
@@ -0,0 +1,88 @@
...
+                        return new URL("file:" + path);
+                    } catch (MalformedURLException e) {
+                        throw new RuntimeException(e);
+                    }
+                })
+                .collect(Collectors.toList());
+    }
+
+    private static Path javaBinPath() {
+        String javaHome = Preconditions.checkNotNull(System.getenv("JAVA_HOME"), "JAVA_HOME not set");
```
carterkozak: I think we want the java-provided `System.getProperty("java.home")` here

palantir-java-format-jre-bootstrap/src/main/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterService.java
fawind: That's true, updated!

palantir-java-format-jre-bootstrap/src/test/java/com/palantir/java/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
fawind: Updated!

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -37,7 +37,7 @@ private PalantirJavaFormatStep() {}
     /** Creates a step which formats everything - code, import order, and unused imports. */
     public static FormatterStep create(Configuration palantirJavaFormat, JavaFormatExtension extension) {
         ensureImplementationNotDirectlyLoadable();
-        Supplier<FormatterService> memoizedService = extension::serviceLoad;
+        Supplier<FormatterService> memoizedService = extension::loadFormatterService;
         return FormatterStep.createLazy(
                 NAME, () -> new State(palantirJavaFormat.getFiles(), memoizedService), State::createFormat);
```
fawind: The gradle integration is very slow because spotless invokes one formatter command per file which will now start a new jvm each. On a medium-sized project (205 java file), formatting all files took ~2 mins. Fortunately files are cached and not reformatted if they don't change.

One solution would be to keep a JVM hot and use this for all formatter invocations. However I would do this as a follow up PR to keep this one smaller. For now, the slow bootstrapping formatter is only used when using Java 16+ (can also make it opt-in to start with).

carterkozak (approved):

fawind:
Let me try deving with the intellij Plugin for a day to see how the latency is.

idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
```diff
@@ -0,0 +1,173 @@
...
+    FormatterService get(Project project, PalantirJavaFormatSettings settings) {
+        return implementationCache.get(new FormatterCacheKey(
+                project, settings.getImplementationClassPath(), settings.injectedVersionIsOutdated()));
+    }
+
+    private static FormatterService createFormatter(FormatterCacheKey cacheKey) {
+        List<URL> implementationUrls =
+                getImplementationUrls(cacheKey.implementationClassPath, cacheKey.useBundledImplementation);
+        Path jdkPath = getJdkPath(cacheKey.project);
+        Integer jdkVersion = getSdkVersion(cacheKey.project);
```
fawind: I thought about using the "same-jvm" formatter for older jvm version but given that in intellij we need the bootstrapping formatter for all projects using Java 12+ it would rather just have one code path.

Also the additional overhead shouldn't be as noticeable in intellij as the usual workflow is to just re-format the current file instead of all files in the project.

carterkozak: > I thought about using the "same-jvm" formatter for older jvm version but given that in intellij we need the bootstrapping formatter for all projects using Java 12+ it would rather just have one code path.

It might be helpful to avoid introducing risk/bugs for existing workflows, to begin with anyhow

fawind: I was thinking that ideally we would already use the bootstrapping formatter for Java 15, which basically includes all internal users at this point.

But given that the formatter didn't support new language features on Java 15 so far, we can start defensively with only enabling the bootstrapping formatter for Java 16+ where it would fail to run. Can enable it for Java 15 in a FLUP PR once this has soaked for a while.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
fawind: We are running into similar problems with spotless in gradle-baseline (see https://github.com/diffplug/spotless/issues/834). As this plugin is more integrated with Gradle, its not super easy to bootstrap this part to a JVM. While waiting for an upstream fix, we will probably use [this](https://github.com/diffplug/spotless/issues/834#issuecomment-819118761) workaround and add the export flags to the `gradle.properties` for the first rollout of Java 17. Based on that, I'll let the pjf gradle integration use the old codepath and not use a bootstrapped jvm.

fawind:
Reverted the changes to the gradle plugin and fixed the case where we would surface errors of the formatter failing on invalid java syntax to the Intellij UI (we now silently ignore these).

Will cut a release now. This effectively just changed the codepath of the Intellij Plugin when using Java 16+ as project SDK.

PR #583, opened 2021-11-18
Fix deadlock trying to format large input

fawind:
## Before this PR
FLUP on https://github.com/palantir/palantir-java-format/pull/582

We had a bug that would result in a deadlock when trying to format large input files using the bootstrapping formatter.

We would wait for the process to finish before draining stdout. For large inputs, this can result in the situation where the process is waiting for availability to write more data to stdout while we wait for the process to finish before draining stdout and providing new availability. As a result the process would indefinitely hang and e.g. freeze Intellij.

## After this PR

Make sure to consume the stdout before waiting for the process to finish.

==COMMIT_MSG==
Fix deadlock trying to format large input
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format-jdk-bootstrap/src/test/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
+++ b/palantir-java-format-jdk-bootstrap/src/test/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
@@ -54,6 +54,22 @@ final class BootstrappingFormatterServiceTest {
         assertThat(formatted).isEqualTo(expectedOutput);
     }
 
+    @Test
+    void can_format_large_input_file() {
+        String input = "class A {\n"
+                + "  void f(){\n"
+                + "    System.out.println(\"Test output\");\n".repeat(2000)
+                + "  }\n"
+                + "}\n";
+
+        ImmutableList<Replacement> replacements =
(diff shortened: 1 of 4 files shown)


palantir-java-format-jdk-bootstrap/src/test/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
```diff
@@ -54,6 +54,22 @@ void can_format_full_file() {
         assertThat(formatted).isEqualTo(expectedOutput);
     }
 
+    @Test
+    void can_format_large_input_file() {
```
fawind, line 58: This test would hang before adding the fix

palantir-java-format-jdk-bootstrap/src/test/java/com/palantir/javaformat/bootstrap/BootstrappingFormatterServiceTest.java
```diff
@@ -54,6 +54,22 @@ void can_format_full_file() {
...
+        String input = "class A {\n"
+                + "  void f(){\n"
+                + "    System.out.println(\"Test output\");\n".repeat(2000)
```
carterkozak, line 61: metaprogramming :-)

carterkozak (approved):

PR #619, opened 2022-01-12
Fix the intellij plugin when the bundled formatter is used

carterkozak:
==COMMIT_MSG==
Fix the intellij plugin when the bundled formatter is used
==COMMIT_MSG==

### DIFF
--- a/idea-plugin/build.gradle
+++ b/idea-plugin/build.gradle
@@ -19,8 +19,9 @@ apply plugin: 'org.jetbrains.intellij'
 
 apply plugin: 'com.palantir.external-publish-jar'
 
+def name = "palantir-java-format"
 intellij {
-    pluginName = "palantir-java-format"
+    pluginName = name
     updateSinceUntilBuild = true
     version = "IU-202.6397.94"
     plugins = ['java']
@@ -73,8 +74,8 @@ tasks.withType(org.jetbrains.intellij.tasks.PrepareSandboxTask) {
     dependsOn configurations.runtimeClasspath
 
     // Also pack the formatter in its own directory
-    into("${pluginName}/impl") {
-        from configurations.formatter
+    into("${name}/impl") {
+        from configurations.formatter.filter { !configurations.runtimeClasspath.contains(it) }
     }
 }
 
--- a/idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
+++ b/idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
@@ -83,13 +83,9 @@ final class FormatterProvider {
 
     private static List<Path> getBundledImplementationUrls() {
         // Load from the jars bundled with the plugin.
-        Path pluginPath = PalantirCodeStyleManager.PLUGIN.getPluginPath();
-        Path implDir = pluginPath.resolve("impl");
-        // Directory appears to have been renamed from 'impl' to 'lib' in recent idea releases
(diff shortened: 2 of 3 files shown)


idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
```diff
@@ -83,13 +83,9 @@ private static FormatterService createFormatter(FormatterCacheKey cacheKey) {
...
-        Path pluginPath = PalantirCodeStyleManager.PLUGIN.getPluginPath();
-        Path implDir = pluginPath.resolve("impl");
-        // Directory appears to have been renamed from 'impl' to 'lib' in recent idea releases
-        Path pathToUse = Files.isDirectory(implDir) ? implDir : pluginPath.resolve("lib");
-
-        log.debug("Using palantir-java-format implementation bundled with plugin: {}", pathToUse);
-        return listDirAsUrlsUnchecked(pathToUse);
+        Path implDir = PalantirCodeStyleManager.PLUGIN.getPath().toPath().resolve("impl");
+        log.debug("Using palantir-java-format implementation bundled with plugin: {}", implDir);
+        return listDirAsUrlsUnchecked(implDir);
```
carterkozak, line 88: reverts a change that went in yesterday, which didn't do anything

idea-plugin/build.gradle
```diff
@@ -73,8 +74,8 @@ tasks.withType(org.jetbrains.intellij.tasks.PrepareSandboxTask) {
     dependsOn configurations.runtimeClasspath
 
     // Also pack the formatter in its own directory
-    into("${pluginName}/impl") {
```
carterkozak, line 76: This variable was not resolved, so the current version was no longer included in our plugin!

idea-plugin/build.gradle
```diff
@@ -73,8 +74,8 @@ tasks.withType(org.jetbrains.intellij.tasks.PrepareSandboxTask) {
...
-        from configurations.formatter
+    into("${name}/impl") {
+        from configurations.formatter.filter { !configurations.runtimeClasspath.contains(it) }
```
carterkozak, line 78: @fawind you removed this `.filter { !configurations.runtimeClasspath.contains(it) }` in #562, do you recall why? It seems beneficial to dedup jars, no?

idea-plugin/build.gradle
fawind: Nice find! Interesting that this worked before!

carterkozak: I'm not entirely sure when it broke since it only failed when the idea plugin used a newer version than the repo

idea-plugin/build.gradle
fawind: If I remember correctly, I ran into issues where dependencies of the formatter where missing when loading the formatter from the impl directory.

Not sure if we have integration tests for this. Can quickly try if the intellij bundled formatter still works after this. Wouldn't be surprised if we can re-add this. I think me removing this filter was mostly a blind guess in trying to fix those issues.

carterkozak: I've verified that it works locally. I was worried that I was testing the wrong path, so I renamed the `impl` directory, restarted my IDE, and it failed to find the formatter impl

fawind (approved):

idea-plugin/build.gradle
carterkozak: I need to revalidate using jdk16+ targets though

fawind: Arg, sorry for the merge then!

PR #707, opened 2022-04-13
Break on lambda argument in method chain

gatesn:
## Before this PR
Chained calls involving lambda expressions don't line break, which conflicts with checkstyle (at least the gradle-baseline checkstyle config)

```java
class PalantirLambdaInliningPrefersBreak {
    void foo() {
        return hello.read(txn -> {
                    doSomeWork();
                    doSomeMoreWork();
                })
                .chainedCall(() -> {
                    doSomeWork();
                });
    }
}

```

## After this PR

```java
class PalantirLambdaInliningPrefersBreak {
    void foo() {
        return hello
                .read(txn -> {
                    doSomeWork();
                    doSomeMoreWork();
                })
                .chainedCall(() -> {
                    doSomeWork();
                });
    }
}
```

## Possible downsides?
🤷‍♂️  I've also added a test to show that non-chained lambda calls don't include a line break

### DIFF
--- a/changelog/@unreleased/pr-707.v2.yml
+++ b/changelog/@unreleased/pr-707.v2.yml
@@ -1,0 +1,16 @@
+type: improvement
+improvement:
+  description: "Break on lambda argument in method chain\n\n## Before this PR\nChained
+    calls involving lambda expressions don't line break, which conflicts with checkstyle
+    (at least the gradle-baseline checkstyle config)\n\n```java\nclass PalantirLambdaInliningPrefersBreak
+    {\n    void foo() {\n        return hello.read(txn -> {\n                    doSomeWork();\n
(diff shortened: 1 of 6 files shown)


robert3005 (approved):

iamdanfox:
@gatesn what was the specific checkstyle thing that was objecting here? I've never encountered this error personally

robert3005:
you can see a test case - it's a chained lambda invocation (with multi statement lambdas) of variable name that's <8 characters

iamdanfox:
DW nick slacked me. My question was because this PR said it was fixing a checkstyle rule, but didn't specific _which_ checkstyle rule. The screenshot showed `Checkstyle: 'block' child has incorrect indentation level 20, expected level should be 12/`

PR #708, opened 2022-04-13
Only apply line break to lambdas with statements

gatesn:
## Before this PR
Previous PR #707 was a little aggressive and inserted line breaks for non-statement lambdas.

## After this PR
Only multi-line lambdas cause a line-break in the method chain.


robert3005:
This does unify the behaviour for short and long identifiers and no longer produces extraneous breaks in case of expression lambdas

robert3005 (approved):

PR #728, opened 2022-04-28
Fix lambda compactness regression

carterkozak:
This reverts #708 and #707 which made expression lambdas less
readable.

==COMMIT_MSG==
Fix lambda compactness regression
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-728.v2.yml
+++ b/changelog/@unreleased/pr-728.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Fix lambda compactness regression
+  links:
+  - https://github.com/palantir/palantir-java-format/pull/728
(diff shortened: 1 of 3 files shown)


CRogers (approved):

robert3005:
This causes conflicts with checkstyle rules in https://github.com/palantir/gradle-baseline/ :/

carterkozak:
I verified that the test output is compatible with baseline checkstyle configuration

robert3005:
ok, I have an actual repro. The lambda has to be more > 10 lines. Will make a pr

robert3005:
after digging more the included test does not pass it. The thing that it boils down to is https://github.com/palantir/gradle-baseline/blob/2a485956a20e7f2ae9a9ddfa689255e347aa96a9/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineConfig.java#L105-L141 and our detection not being fully robust. So... my proposal here is to kill that logic I linked and make that the default checkstyle?

gatesn:
@carterkozak do you have an example of a less readable case? 

Happy with the revert if we fix the checkstyle plugin config issue, but I also thought it looked a bit prettier!

PR #731, opened 2022-04-28
Improvement: Publish gradle plugin to gradle plugin portal

robert3005:
## After this PR

==COMMIT_MSG==
Publish gradle plugin to gradle plugin portal
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,13 +5,14 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.palantir.gradle.externalpublish:gradle-external-publish-plugin:1.11.0'
-        classpath 'com.palantir.javaformat:gradle-palantir-java-format:2.23.0'
-        classpath 'com.palantir.gradle.revapi:gradle-revapi:1.7.0'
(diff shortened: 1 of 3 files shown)


iamdanfox:
IIRC the `:palantir-java-format-spi` was a blocker to successfully publishing to the gradle plugin portal... dan had this originally: https://github.com/palantir/palantir-java-format/pull/24

robert3005:
I really care that the gradle plugin markers are published so I could just turn on automatedpublishing and not do the rest

robert3005:
eh, this should just work as the spi is published as a library

iamdanfox:
Actually I think the problem before was that using the `plugins {}` syntax had a hard-coded list of repositories that it would try to load jars from, and our old bintray thing was obviously not one of them... given that we're now publishing to MavenCentral, maybe this is no longer a problem??

Think this is worth a shot, but will defer the final +1 to a current infra team member.

robert3005:
you can always add a repo under `pluginManagement` in settings.gradle

carterkozak (approved):
lgtm

PR #874, opened 2023-05-22
List our IntelliJ compatibility correctly

CRogers:
## Before this PR
As per support email reply from Jetbrains, our 2.30.0 release is not being approved as it is incompatible with builds lower than 193. We can fix this temporarily by changing it through the UI, or we can fix it forever by modifying the `plugin.xml` included with the plugin.

## After this PR
==COMMIT_MSG==
Update intellij compatibility range in IntelliJ plugin `plugin.xml` metadata file.
==COMMIT_MSG==

### DIFF
--- a/idea-plugin/src/main/resources/META-INF/plugin.xml
+++ b/idea-plugin/src/main/resources/META-INF/plugin.xml
@@ -9,6 +9,7 @@
        See https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html#modules-specific-to-functionality
    -->
   <depends>com.intellij.modules.java</depends>
+  <idea-version since-build="193"/>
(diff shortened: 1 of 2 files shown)


idea-plugin/src/main/resources/META-INF/plugin.xml
```diff
@@ -9,6 +9,7 @@
        See https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html#modules-specific-to-functionality
    -->
   <depends>com.intellij.modules.java</depends>
+  <idea-version since-build="193"/>
```
fawind, line 12: For my reference: That's the 2019.3 release.

fawind (approved):

PR #875, opened 2023-05-22
Update IntelliJ compatibility bounds correctly

CRogers:
## Before this PR
In #874, I tried to update the minimum intellij version this is listed to work with. However, [it didn't work](https://plugins.jetbrains.com/plugin/13180-palantir-java-format/versions/stable/337985) as the `patchPluginXml` task changes the `plugin.xml` file.

## After this PR
Use the correct intellij version bound.

==COMMIT_MSG==
Update intellij compatibility range in IntelliJ plugin `plugin.xml` metadata file, but hopefully correctly this time.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-875.v2.yml
+++ b/changelog/@unreleased/pr-875.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Update intellij compatibility range in IntelliJ plugin `plugin.xml`
+    metadata file, but hopefully correctly this time.
+  links:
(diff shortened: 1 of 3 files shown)


CRogers:
I ran `./gradlew buildPlugin` then checked that the plugin jar contains this in it's `plugin.xml`:

```xml
<idea-version since-build="193" />
```

by running `vim idea-plugin/build/libs/idea-plugin-2.31.0-1-gfdceb49.jar`

fawind (approved):

PR #924, opened 2023-09-07
Remove outdated docs

Nifdee:
Save Actions is no longer compatible with latest IntelliJ, so PJF now uses IntelliJ’s native “Actions on Save” functionality instead.


fawind (approved):
Thanks!

Issue #833, opened 2023-01-24
What is the easiest way to use this in Eclipse?


ralienpp:
I pitched the use of `palantir-java-format` to my colleagues, but now I'm facing the issue of on-boarding people who use Eclipse. After exploring the available materials and issue discussions, I understand that there's no way to do it with a few clicks or by installing a plugin.

Thus, my question is: _what is the path of minimum resistance for using `palantir-java-format` with the Eclipse IDE?_.

Git hooks come to mind, one could invoke `mvn spotless:apply` when a commit is made. This is an IDE-agnostic approach, and I like it. However, it disregards Eclipse's formatting tools and keyboard shortcuts - which means that people won't see the auto-formatted code until they commit. Moreover, people's workflow that they developed for ages (write code, format it) will potentially be disrupted because there is no feedback loop.

Is there a way to tell Eclipse to run a given command when a file is saved? What I found so far in `Project properties / Java editor / Save actions` is not at all CLI-friendly, it can perform some actions from a set of possibilities that Eclipse defined, but there's no way to delegate this process to an arbitrary command.

What are the best practices when it comes to `palantir-java-format` and Eclipse?

PR #949, opened 2023-11-09
Fixes the Eclipse Plugin

ppkarwasz:
This PR fixes the Eclipse Plugin by backporting some code from Google Java Format and solving some configuration problems.

The plugin uses the Eclipse Tycho Maven plugin to build, although it can probably be easily adapted to Gradle.

It was tested on Eclipse 2022-09 under JDK 17. Under JDK 17 the plugin requires additional setup, which is described in `README.md`.

The plugin requires a JRE 11 to run (same requirement as Eclipse 2020-09 or later).

## Before this PR

The Eclipse Plugin didn't even compile.

## After this PR

There should a functional Eclipse Plugin.


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @ppkarwasz! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ppkarwasz:
@carterkozak, could you take a look at this?

ppkarwasz:
Fixes #833

carterkozak:
I’ll take a closer look when I have a chance, I’d really like to use a gradle plugin to build this plugin so we can publish the compiled jar to maven central in the build process. I’m not sure how feasible that would be, though.

ppkarwasz:
> I’ll take a closer look when I have a chance, I’d really like to use a gradle plugin to build this plugin so we can publish the compiled jar to maven central in the build process. I’m not sure how feasible that would be, though.

The only complex operation in the build is downloading Eclipse artifacts from a P2 repository. I think that [diffplug/goomph](https://github.com/diffplug/goomph) could help with that, which redirects to the new [equodev/equo-ide](https://github.com/equodev/equo-ide/tree/main/plugin-gradle#user-plugins) project.

My Gradle skills are rather limited, but I'll try to come up with something.

In the current state the PR could still be useful, since it allows users to compile the plugin themselves.

ppkarwasz:
I managed to switch to a Gradle build.

eclipse_plugin/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target/
+/.mvn/wrapper/maven-wrapper.jar
```
carterkozak, line 2: I think this file can be deleted now that we're using gradle

carterkozak: I can take care of this, there are a few gradle-shaped things I plan to experiment with as well (I'd like to pull the list of dependencies listed in the manifest dynamically, and wire up plugins for maven publishing)

carterkozak:
> I managed to switch to a Gradle build.

Impressive, thanks!

eclipse_plugin/.gitignore
ppkarwasz: Go ahead, my Gradle knowledge is quite limited, so I limited myself to a static `MANIFEST.MF`. However the `Bundle-Classpath` can be certainly generated dynamically, so that the deps keep their original names.

**Remark**: some dependencies like `guava` or `jackson` are already in the official Eclipse P2 repositories. I just don't know how to force Eclipse to install additional bundles, probably this would require an "Eclipse feature" distributed in a P2 repository.

rombert:
Any plans to see this merged?

carterkozak (approved):

Issue #977, opened 2024-01-08
Upgrade Gradle to version 8.5 fails


matsev:
## What happened?
Upgrading Gradle wrapper to version 8.5 does not currently work:

```bash
./gradlew wrapper --gradle-version 8.5

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/mseverson/Projects/palantir/palantir-java-format/gradle-palantir-java-format/build.gradle' line: 27

* What went wrong:
A problem occurred evaluating project ':gradle-palantir-java-format'.
> Adding a Configuration as a dependency is no longer allowed as of Gradle 8.0.

```
This is documented in the [Configurations no longer allowed as Dependencies](https://docs.gradle.org/8.5/userguide/upgrading_version_7.html#configurations_no_longer_allowed_as_dependencies) in the Gradle user guide:

> Adding a Configuration as a dependency in the dependencies DSL block, or programmatically using the DependencyHandler classes' doAdd(Configuration, Object, Closure) method, is no longer allowed and will fail with an exception. To replicate many aspects of this behavior, extend configurations using the extendsFrom(Configuration) method on Configuration instead.

## What did you want to happen?

Upgrading Gradle wrapper should work

## Why this is relevant

The [Gradle Compatibility Matrix](https://docs.gradle.org/current/userguide/compatibility.html) states that:
- Gradle version 8.4 is required for compiling and testing Java 21 projects
- Gradle version 8.5 is required for running Gradle on Java 21 JVM

Consequently, upgrading Gradle is a blocker for Java 21 related tickets, e.g. #931, #933, #934, #952, #966

koppor:
A workaround is to use [Gradle's toolchain feature](https://docs.gradle.org/current/userguide/toolchains.html), which relatively simple to implement: https://github.com/palantir/palantir-java-format/pull/978 - and enables parallel work on an update of Gradle and working on Java 21 related issues.

matsev:
@koppor I just tried to build locally using the Gradle toolchain config that is currently provided in #978. Regrettably my build fails with a similar stack trace as the one [triggered on the build server](https://circleci.com/gh/palantir/palantir-java-format/14357):

```bash
> Task :palantir-java-format-spi:compileJava FAILED
error: An unhandled exception was thrown by the Error Prone static analysis plugin.
     Please report this at https://github.com/google/error-prone/issues/new and include the following:
  
     error-prone version: 2.19.1
     BugPattern: (see stack trace)
[...]

PR #978, opened 2024-01-09
Switch to Java 21 for compilation

koppor:
## Before this PR

palantir-java-format relied on Java 11 for compilation.

Gradle update does not work. See https://github.com/palantir/palantir-java-format/issues/977.

## After this PR

palantir-java-format relias on Java 21 for compilation, but sticks with Gradle 7.6.2.

This enables working on Java 21 related tickets.

## Possible downsides?

Java 21 might be a too strong requirement for some "legacy" code. One could release `3.0.0` and say, hey, 2.x supports up to Java 20, 3.x supports 21+.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ aliases:
 jobs:
 
   check:
-    docker: [{ image: 'cimg/openjdk:17.0.1-node' }]
+    docker: [{ image: 'cimg/openjdk:11.0.21-node' }]
     resource_class: large
     environment:
       CIRCLE_TEST_REPORTS: /home/circleci/junit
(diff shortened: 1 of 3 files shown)


koppor:
This damn error prone is not compatible with Java 21. And I don't know how to remove it from the build configuration.

koppor:
I have following error now:

    error: exporting a package from system module jdk.compiler is not allowed with --release

It is not possible to remove `--release` in gradle 7.x. - cf. https://github.com/gradle/gradle/pull/22449.

stackoverflow also wasn't of help. Cf. https://stackoverflow.com/q/70477692/873282

Thus closing this.

Issue #724, opened 2022-04-26
Please expose option to format JavaDoc


ChristianCiach:
In https://github.com/palantir/palantir-java-format/pull/152 you've disabled the formatting of JavaDoc by hardcoding a "false" value.

Google-Java-Format provides a command-line switch (and API) to enable/disable the formatting of JavaDoc. I would like you to do the same :)

carterkozak:
This sounds reasonable to me, I've been meaning to add support for this feature to support our javapoet formatter, [Goethe](https://github.com/palantir/goethe).

ChristianCiach:
That being said, I am not sure about the API while keeping backwards compatibility.

In my eclipse-plugin I am currently using the ServiceLoader to retrieve an instance of `com.palantir.javaformat.java.FormatterService`, which is an interface with these two methods:

```java
    ImmutableList<Replacement> getFormatReplacements(String input, Collection<Range<Integer>> ranges)
            throws FormatterException;

    String formatSourceReflowStringsAndFixImports(String input) throws FormatterException;
```

For keeping backwards compatibility maybe we should change these two methods to "default" methods to delegate to new methods that take an additional parameter `boolean formatJavaDoc`, like so:

```java
    ImmutableList<Replacement> getFormatReplacements(String input, Collection<Range<Integer>> ranges, boolean formatJavaDoc);

    String formatSourceReflowStringsAndFixImports(String input, boolean formatJavaDoc) throws FormatterException;

    default ImmutableList<Replacement> getFormatReplacements(String input, Collection<Range<Integer>> ranges)
            throws FormatterException {
        return getFormatReplacements(input, ranges, false);
    }

    default String formatSourceReflowStringsAndFixImports(String input) throws FormatterException {
        return formatSourceReflowStringsAndFixImports(input, false);
    }
```

This way we could keep the old methods (maybe deprecating them).

q3769:
#944

ChristianCiach:
Thanks for notifying me. Closing this issue as done.

q3769-patientpoint:
> Thanks for notifying me. Closing this issue as done.

Sorry why is this one closed? It's not done.

There are couple of related issues: One is for Geothe, which is done; the other is for IntelliJ/idea plugin, which will most likely be automatically done if this current issue is worked on and completed. 

My understanding of the current issue/request is to provide an option to enable Javadoc formatting on the CLI or core level, which is not done.

ChristianCiach:
[...]

PR #986, opened 2024-01-21
Add option to IDEA plugin (and CLI) to format JavaDoc #724

blutorange:
As discussed in #724 (as I understand it), I added an option to the IDEA plugin to enable formattting JavaDoc comments.

* This adds an option to the IDEA plugin (checkbox) to format JavaDoc comments. It is disabled by default.
* As the IDEA plugin (sometimes) uses the bootstrapping formatter service, which in turn uses the command line tool, this also add an option `--format-javadoc` to the command line tool.
* To pass the options to the FormatterService loaded via service provider, I added a method `withOptions` to the service provided interface, which returns a new (immutable) instance with the updated options. Compared with the suggestion from the issue comment (https://github.com/palantir/palantir-java-format/issues/724#issuecomment-1109804294), this (a) does not require deprecating methods and (b) the options can be set independently of where the formatter is actually used.
* Since the infrastructure to pass options to the formatter now exists, we could also add more options to the code style select configuration in the IDEA plugin, but I'm not sure if that makes sense (I want to use Palantir, after all).
* The issue also mentioned API compatibility. The service provider interface is compatible, most other classes and methods are internal. But if somebody more familar with the code base could take another look, that would be great.
* I also added a few tests for the new option.

## Possible downsides?

* This adds an option that affects the formatter output, and Palantir strives to be an opinionated formatter without any configuration. As discussed in the issue, though, adding an option for fomatting JavaDoc should be ok.

image


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @blutorange! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

blutorange:
@schlosna  As far as I can see, the changelog needs to be generated by somebody with the proper permissions for this repo?

blutorange:
Is there anything I still need to do before this PR can be considered for review?

walles:
This branch has conflicts that must be resolved.

blutorange:
Thanks for notifying me! I resolved the conflicts in case anybody wants to take a look at getting this merged.

blutorange:
Is there any chance of getting this merged? Are you open to adding this option to the plugin or should I close this?

blutorange:
Seems there are new merge conflict now that this PR has been around for so long. I can take a look at it, if there's any chance / interest in getting this merged? @crogoz

ppkarwasz:
I’d really love to see this merged. I’m a big fan of the Palantir formatter: one of its biggest wins is that it ends discussions about style and formatting. With something like the Eclipse formatter, we’d still be arguing over trivial whitespace or line break options.

Unfortunately Javadoc-heavy PRs are still especially prone to bike-shedding over small details. Having this option would reduce noise, let contributors focus on real issues, and keep the style consistent without endless debates.

awa-xima:
I can only second this. I write a lot of JavaDoc and my biggest pain is having to handle line breaks manually, which becomes annoying especially when you need to edit existing JavaDoc. We use spotless-maven-plugin to run palantir-java-format on commit, and I already added JavaDoc formatting to the spotless-maven-plugin. So at worst, I can just ignore line breaks and it will get formatted once I commit. But it would still be nice to see the formatting immediately in the IDE.

There are new releases all the time, but most seem to be auto releases. The most recent commits in this repo seem to be from @crogoz ? I'll take a look at fixing the merge conflicts if you'd be ok with adding this option?

PR #988, opened 2024-01-21
Enable Nullaway

### LINKED ISSUE #987: Apply baseline's NullAway on this repo
## What happened?
At https://github.com/palantir/palantir-java-format/pull/980/files#r1451076860 I accidentally wrote code returning a `null` from a method with `OptionalInt` return type.  Ideally the [baseline-null-away](https://github.com/palantir/gradle-baseline/tree/develop/baseline-null-away) plugin would be applied, to automatically flag issues like this.

ash211:
Fixes https://github.com/palantir/palantir-java-format/issues/987

## Before this PR
Nullaway is not enabled, and null handling isn't verified by that plugin.

## After this PR
==COMMIT_MSG==
Enable Nullaway
==COMMIT_MSG==

Fix flagged issues by adding `@Nullable` and handling nulls better with `Optional.ofNullable`

## Possible downsides?
NullAway might fail future PRs more often; it can be fiddly in requirements that tend to be largely unimportant in practice (like that Exception messages can be null).


palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
```diff
@@ -130,7 +132,7 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
             }
         }
 
-        for (Map.Entry<Path, Future<String>> result : results.entrySet()) {
+        for (Entry<Path, Future<String>> result : results.entrySet()) {
```
schlosna: Was this intentional? I'm surprised this didn't trigger error-prone [BadImport](https://errorprone.info/bugpattern/BadImport) 

```suggestion
        for (Map.Entry<Path, Future<String>> result : results.entrySet()) {
```

palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
```diff
@@ -215,10 +220,10 @@ public static CommandLineOptions processArgs(String... args) throws UsageExcepti
         try {
             parameters = CommandLineOptionsParser.parse(Arrays.asList(args));
         } catch (IllegalArgumentException e) {
-            throw new UsageException(e.getMessage());
+            throw new UsageException(Optional.ofNullable(e.getMessage()).orElse("null"));
```
schlosna: I don't think we need to change this, but in other types of hot paths [`java.until.Objects.requireNonNullElse`](https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.base/java/util/Objects.html#requireNonNullElse(T,T)) can avoid the Optional allocation

```suggestion
            throw new UsageException(Objects. requireNonNullElse(e.getMessage(), "null"));
```

schlosna (dismissed):
Overall looks good, thanks for the cleanup!

palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
ash211: `"Entry"` isn't on BadImport's list: https://github.com/google/error-prone/blob/4165ecdc3d11a41b989d08e324de6fbc944e95ae/core/src/main/java/com/google/errorprone/bugpatterns/BadImport.java#L73-L84

Maybe it should be?

Fixed

palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
ash211: Thanks for sharing -- I've needed this before and its good to know a way to do this without the allocation

palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
schlosna: Wow, looks like Entry was removed from the BadImport list 3 years ago, but my brain has already been conditioned to qualify Entry.

https://github.com/google/error-prone/commit/38a0e4b79ca04b20543ac8051b074c91def841f9

schlosna (approved):

PR #989, opened 2024-01-22
Improvement: Add instrumentation to allow warning on files that take long to format

MTelling:
## Before this PR
We have noticed the formatter becomes really slow on functions that nest a lot, causing some files to take +10 seconds to format. 
A simple refactor can often fix these issues but it can be hard to identify and enforce that files are kept in a shape that formatter formats quickly.

This PR adds instrumentation to file formatting and, if enabled, prints errors to the console for files that exceed the specified threshold. 

## After this PR

==COMMIT_MSG==
Add instrumentation to allow warning on files that take long to format
==COMMIT_MSG==

## Possible downsides?
I'd rather fix the formatter here but that might take a lot more time and in places where we've seen issues, the code could use a refactor anyway (deeply nested code is not very pretty).

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptions.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptions.java
@@ -45,6 +45,7 @@ final class CommandLineOptions {
     private final Optional<String> assumeFilename;
     private final boolean reflowLongStrings;
     private final boolean outputReplacements;
+    private final Optional<Long> warnOnExpensiveFileDurationMillis;
 
     CommandLineOptions(
             ImmutableList<String> files,
@@ -65,7 +66,8 @@ final class CommandLineOptions {
             boolean setExitIfChanged,
             Optional<String> assumeFilename,
             boolean reflowLongStrings,
-            boolean outputReplacements) {
+            boolean outputReplacements,
+            Optional<Long> warnOnExpensiveFileDurationMillis) {
         this.files = files;
         this.inPlace = inPlace;
         this.lines = lines;
@@ -85,6 +87,7 @@ final class CommandLineOptions {
         this.assumeFilename = assumeFilename;
         this.reflowLongStrings = reflowLongStrings;
         this.outputReplacements = outputReplacements;
+        this.warnOnExpensiveFileDurationMillis = warnOnExpensiveFileDurationMillis;
     }
 
     /** The files to format. */
@@ -185,6 +188,11 @@ final class CommandLineOptions {
         return outputReplacements;
     }
 
+    /** Returns the number of millis to allow processing of a single file to take before warning. */
+    Optional<Long> warnOnExpensiveFileDurationMillis() {
+        return warnOnExpensiveFileDurationMillis;
+    }
+
     static Builder builder() {
         return new Builder();
     }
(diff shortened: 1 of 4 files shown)


palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptions.java
```diff
@@ -45,6 +45,7 @@ final class CommandLineOptions {
     private final Optional<String> assumeFilename;
     private final boolean reflowLongStrings;
     private final boolean outputReplacements;
+    private final Optional<Long> warnOnExpensiveFileDurationMillis;
```
MTelling, line 48: I would probably long term want to be able to set a `failIfDurationExceeded` option as well. Ideas for if this flag should be altered or if it's just another config?

palantir-java-format/src/main/java/com/palantir/javaformat/java/InstrumentedFormatFileCallable.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public InstrumentedFormatFileCallable(FormatFileCallable delegate) {
+        this.delegate = delegate;
+    }
+
+    @Override
+    public FormatFileResult call() throws Exception {
+        long start = System.currentTimeMillis();
+        String result = delegate.call();
+        Duration duration = Duration.ofMillis(System.currentTimeMillis() - start);
```
ash211, line 39: if you have Guava on the classpath already, use `Stopwatch`?

palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptions.java
```diff
@@ -185,6 +188,11 @@ boolean outputReplacements() {
         return outputReplacements;
     }
 
+    /** Returns the number of millis to allow processing of a single file to take before warning. */
+    Optional<Long> warnOnExpensiveFileDurationMillis() {
```
ash211, line 192: Can this be `Optional<Duration>` ?

palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptions.java
ash211: just another config, imo

palantir-java-format/src/main/java/com/palantir/javaformat/java/Main.java
```diff
@@ -177,6 +183,16 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
         return allOk ? 0 : 1;
     }
 
+    private void warnIfDurationExceedsThreshold(CommandLineOptions parameters, Path path, FormatFileResult fileResult) {
+        if (parameters.warnOnExpensiveFileDurationMillis().isPresent()
+                && fileResult.duration().toMillis()
+                        > parameters.warnOnExpensiveFileDurationMillis().get()) {
+            errWriter.println(path + ": took " + fileResult.duration().toMillis() + "ms to format, "
```
ash211, line 190: nit: put `WARN` in this message, so it shows up when people Cmd+F for warnings (which log but do not fail builds)

ash211 (commented):
This would be helpful for a project I'm working on where we are spending nearly as much time on spotless as we are on compilation!

- 37m 15.661s on Spotless
- 45m 5.492s on Compile

PR #1029, opened 2024-03-13
Avoid computing expensive debug info for all checkState() calls

ash211:
## Before this PR
`level.representation(State.startingState())` is always calculated, regardless of if the `checkState()` check fails.  This is expensive to compute, so we should avoid doing it unless necessary.

Screenshot 2024-03-13 at 12 33 14 AM

## After this PR
==COMMIT_MSG==
Avoid computing expensive debug info for all checkState() calls
==COMMIT_MSG==

## Possible downsides?
None known


fawind (approved):
Thanks!

PR #1037, opened 2024-03-15
Fix error message link for filing bugs

ash211:
## Before this PR
Users see this error message and are urged to file bugs on the wrong repo.  Palantir forked this repo significantly enough that we probably shouldn't be sending bug reports Google's way.

## After this PR
==COMMIT_MSG==
Fix error message link for filing bugs
==COMMIT_MSG==

## Possible downsides?
None known


CRogers:
:+1:

PR #1069, opened 2024-04-05
Fix: Properly format switch expressions

### LINKED ISSUE #1025: Poor formatting around expression switches
## What happened?

Pretty interesting, seems to only happen when the expression is assigned to a variable that was already declared.

```
    public void test1(int y) {
        int x =
                switch (y) {
                    case 1 -> 1;
                    case 2 -> throw new IllegalArgumentException();
                    default -> throw new IllegalStateException();
                };
    }

    public void test2(int y) {
        int x;
        x = switch (y) {
            case 1 -> 1;
            case 2 -> throw new IllegalArgumentException();
            default -> throw new IllegalStateException();};
    }
```

Probably this?

```
    public void test2(int y) {
        int x;
        x = switch (y) {
            case 1 -> 1;
            case 2 -> throw new IllegalArgumentException();
            default -> throw new IllegalStateException();
        };
    }
```

johnhany97:
## Before this PR
Fixes https://github.com/palantir/palantir-java-format/issues/1025 where before this change, we'd format switch expressions used in an assignment to yield the following:
```java
    public void test2(int y) {
        int x;
        x = switch (y) {
            case 1 -> 1;
            case 2 -> throw new IllegalArgumentException();
            default -> throw new IllegalStateException();};
    }
```

## After this PR

==COMMIT_MSG==
Properly format switch expressions
==COMMIT_MSG==
and now we generate this:
```java
    public void test2(int y) {
        int x;
        x = switch (y) {
            case 1 -> 1;
            case 2 -> throw new IllegalArgumentException();
            default -> throw new IllegalStateException();
        };
    }
```

### DIFF
--- a/gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/FormatDiffTest.java
+++ b/gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/FormatDiffTest.java
@@ -63,6 +63,7 @@ class FormatDiffTest {
         runCommandInRepo("git", "init");
         runCommandInRepo("git", "config", "user.name", "Test User");
         runCommandInRepo("git", "config", "user.email", "test-user@palantir.com");
+        runCommandInRepo("git", "config", "commit.gpgsign", "false");
         runCommandInRepo("git", "commit", "--allow-empty", "-m", "Init");
 
         Path subdir = repo.resolve("subdir");
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.output
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.output
@@ -47,4 +47,22 @@ class ExpressionSwitch {
             default -> false;
         };
     }
+
+    public void test1(int y) {
+        int x =
+                switch (y) {
+                    case 1 -> 1;
+                    case 2 -> throw new IllegalArgumentException();
+                    default -> throw new IllegalStateException();
+                };
+    }
+
+    public void test2(int y) {
+        int x;
+        x = switch (y) {
+            case 1 -> 1;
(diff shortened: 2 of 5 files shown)


gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/FormatDiffTest.java
```diff
@@ -63,6 +63,7 @@ void reformat_a_subpath_of_a_git_directory_for_only_changed_lines() throws IOExc
         runCommandInRepo("git", "init");
         runCommandInRepo("git", "config", "user.name", "Test User");
         runCommandInRepo("git", "config", "user.email", "test-user@palantir.com");
+        runCommandInRepo("git", "config", "commit.gpgsign", "false");
```
johnhany97, line 66: Drive-by improvement: Didn't see it as necessary to sign the commits made by the test

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.output
```diff
@@ -47,4 +47,22 @@ class ExpressionSwitch {
...
         };
     }
+
+    public void test1(int y) {
+        int x =
+                switch (y) {
+                    case 1 -> 1;
+                    case 2 -> throw new IllegalArgumentException();
+                    default -> throw new IllegalStateException();
+                };
```
johnhany97, line 57: Would like to fix this in the future FWIW to instead be:
```java
        int x = switch (y) {
            case 1 -> 1;
            case 2 -> throw new IllegalArgumentException();
            default -> throw new IllegalStateException();
        };
```
But that requires more complicated changes to happen to the `visitAssignment` method which should be left to a separate PR

carterkozak (approved):
tyvm!

PR #1078, opened 2024-04-26
Move the plugin to the FormattingService & ImportOptimizer APIs

### LINKED ISSUE #1068: Formatting shortcut not intercepted in IntelliJ 2024.1
## What happened?

Using 2.40.0 of the IntelliJ plugin on recently released 2024.1, the format shortcut does not seem to be intercepted anymore and the default IntelliJ formatter is run instead of palantir.

The format shortcut should be intercepted on IntelliJ 2024.1

crogoz:
## Before this PR
IntelliJ 2024.1 stops formatting via palantir-java-format when you press the format shortcut key or save (if formatting on actions on save is enabled).

## After this PR
Similar to [this google-java-format commit](https://github.com/google/google-java-format/commit/84b2c9a2ba382ee7a968518b46afb0d6000ada8d?diff=unified&w=0), we move to the new `FormattingService` and `ImportOptimizer` apis rather than doing the replacement of `CodeStyleManager` that we used to do.

Fixes #1068.


idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -0,0 +1,157 @@
...
+        return JavaFileType.INSTANCE.equals(file.getFileType())
+                && PalantirJavaFormatSettings.getInstance(file.getProject()).isEnabled();
+    }
+
+    @Override
+    public @NotNull Set<ImportOptimizer> getImportOptimizers(@NotNull PsiFile file) {
+        Project project = file.getProject();
+        PalantirJavaFormatSettings settings = PalantirJavaFormatSettings.getInstance(project);
+        Optional<FormatterService> formatter = formatterProvider.get(project, settings);
+        return Set.of(new PalantirJavaFormatImportOptimizer(formatter));
```
crogoz, line 78: different from google-java-format: we need to pass the formatter to the Optimizer class, such that we can call the spi implementation

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -0,0 +1,157 @@
...
+                request.onTextReady(formattedText);
+            } catch (FormatterException e) {
+                request.onError(
+                        Notifications.PARSING_ERROR_TITLE,
+                        Notifications.parsingErrorMessage(
+                                request.getContext().getContainingFile().getName()));
+            }
+        }
+
+        public static String applyReplacements(String input, Collection<Replacement> replacementsCollection) {
```
crogoz, line 113: Same implementation as in `Utils.applyReplacements`

idea-plugin/src/test/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+public class PalantirJavaFormatFormattingServiceTest {
+    private JavaCodeInsightTestFixture fixture;
+    private PalantirJavaFormatSettings settings;
+    private DelegatingFormatter delegatingFormatter;
+
+    @BeforeEach
+    public void setUp() throws Exception {
+        TestFixtureBuilder<IdeaProjectTestFixture> projectBuilder = IdeaTestFixtureFactory.getFixtureFactory()
+                .createLightFixtureBuilder(getProjectDescriptor(), getClass().getName());
```
crogoz, line 60: using `.createLightFixtureBuilder(getProjectDescriptor(), getClass().getName());` instead of `.createFixtureBuilder(getClass().getName());` such that we can set up the sdk in: https://github.com/palantir/palantir-java-format/pull/1078/files#diff-bc3ac8002ebba51e0734b16f06fb1ebf8aadbe6c006c247cb80fcd72c877a2f5R88

same for the other test

idea-plugin/src/test/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @AfterEach
+    public void tearDown() throws Exception {
+        fixture.tearDown();
+    }
+
+    @Test
+    public void defaultFormatSettings() throws Exception {
+        String input = Files.readString(
```
crogoz, line 81: Deviates from the google-java-format test. This test only checks the `PalantirJavaFormatFormattingService` workflow.

palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/FormatterService.java
```diff
@@ -48,4 +48,15 @@ ImmutableList<Replacement> getFormatReplacements(String input, Collection<Range<
...
+    /**
+     * Fixes imports (eg. ordering, spacing, and removal of unused import statements).
+     *
+     * @param input the input string
+     * @return the output string
+     * @throws FormatterException if the input string cannot be parsed
+     * @see <a href="https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing">Google Java
+     *     Style Guide - 3.3.3 Import ordering and spacing</a>
+     */
+    String fixImports(String input) throws FormatterException;
```
CRogers, line 57: I think this will cause issues when there is an old version of the Gradle plugin but this new version of IntelliJ plugin exists. The old `BootstrappingFormatterService` will be service loaded and will not have this method, which the JVM will not be happy about.

Making this a default method that returns `Optional<>` would fix this, so we could detect if the `fixImports` method doesn't exist.

However, we also have the opposite problem - when the version of the Gradle plugin is new but the IntelliJ plugin is old, we'll try to load a `BootstrappingFormatterService` that has an extra method that does not exist in the interface - I'm not 100% sure whether this is a break, but I think the JVM is going to be unhappy again. Needs some real testing.

I'm concerned this might have been built in a way that means we can't really change it in a way that's both forwards and backwards compatible.

palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/FormatterService.java
```diff
@@ -48,4 +48,15 @@ ImmutableList<Replacement> getFormatReplacements(String input, Collection<Range<
...
+
+    /**
+     * Fixes imports (eg. ordering, spacing, and removal of unused import statements).
+     *
+     * @param input the input string
+     * @return the output string
+     * @throws FormatterException if the input string cannot be parsed
+     * @see <a href="https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing">Google Java
+     *     Style Guide - 3.3.3 Import ordering and spacing</a>
+     */
```
CRogers: I don't think the javadoc adds much - especially as we don't necessarily use Google's import ordering guidelines.

palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/FormatterService.java
CRogers: The solution is probably to make another interface and "make do" with the previous one - ie formatting and import ordering at the same time possibly with a warning.

idea-plugin/build.gradle
```diff
@@ -23,7 +23,7 @@ def name = "palantir-java-format"
 intellij {
     pluginName = name
     updateSinceUntilBuild = true
-    version = "IU-202.6397.94"
+    version = "2024.1"
```
CRogers, line 26: Will need to check if this is actually true, but I think we need to set this back down to a lower version before we merge - google-java-format [had it at 2021.3](https://github.com/google/google-java-format/commit/84b2c9a2ba382ee7a968518b46afb0d6000ada8d#diff-6c907a8913b20e03404780aeb811db4c391febccde44b69ac482d896864a802aR38).

palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/FormatterService.java
CRogers: I say all this, but testing it out it seems to work somehow in `./gradlew runIde` with older versions of the Gradle plugin - maybe it's fine or there's something I don't understand happening with the classloaders here. I'll give the other way around (old intellij plugin, new gradle plugin) a manual test too.

CRogers: Newer gradle plugin, older intellij plugin also seems to work 🤔

crogoz: > I think this will cause issues when there is an old version of the Gradle plugin but this new version of IntelliJ plugin exists. The old BootstrappingFormatterService will be service loaded and will not have this method, which the JVM will not be happy about.

When I was running the tests using `./gradlew runIde` the newest version of Intelij was running (2024.1). I opened a project that had an old `gradle-palantir-java-format` plugin applied and the formatter worked. The formatter that was picked up was bundled from the palantir-java-plugin. There seems to be this logic [here](https://github.com/palantir/palantir-java-format/blob/2ec339a038e24c78b27509af9f2c7cd20ac798d5/idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java#L114) which gets set up [here](https://github.com/palantir/palantir-java-format/blob/2ec339a038e24c78b27509af9f2c7cd20ac798d5/idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatSettings.java#L91). So from what I understand, the higher version of the formatter will always be picked up. 

> However, we also have the opposite problem - when the version of the Gradle plugin is new but the IntelliJ plugin is old, we'll try to load a BootstrappingFormatterService that has an extra method that does not exist in the interface - I'm not 100% sure whether this is a break, but I think the JVM is going to be unhappy again. Needs some real testing.

Tested the workflow by bumping my Intelij version to latest. (2024.1 - without my fix) and used the locally published `gradle-palantir-java-format` (with the fix). The formatter worked correctly (and I didn't see any errors in the idea.log file)

> The solution is probably to make another interface and "make do" with the previous one - ie formatting and import ordering at the same time possibly with a warning.

Wdyt if I do create a new interface that will only be used by the Intelij plugin ?

idea-plugin/build.gradle
crogoz: when running `./gradlew runIde` the intelij version is picked up from this version. Can leave it at 2021.3 though

palantir-java-format-spi/src/main/java/com/palantir/javaformat/java/FormatterService.java
CRogers: Ah, the using the bundled version rather than the gradle implementation was the bit I was missing - this all makes sense now.

idea-plugin/build.gradle
CRogers: The [docs](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#intellij-extension-version) say:

> The version of the IntelliJ Platform IDE that will be used to build the plugin. 

I think this means the SDK, especially given the errors about `LightProjectDescriptor` not existing in the tests. We should check whether this means you need at least this version of IntelliJ to run it (or perhaps it's fine provided you don't use "new" methods?)

crogoz: > especially given the errors about LightProjectDescriptor not existing in the tests.

yes, I think this is the version used for building & testing the plugin

> least this version of IntelliJ to run it 

I think 'since-build' does this. So we are saying that our plugin is compatible with at least 2021.3 (213)

CRogers: Yeah, I think it's probably fine - the only risk is someone later tries to use some very new method. But I don't expect any work on this intellij plugin needed beyond bugfixes...

CRogers: Yeah, the plugin built using 2024.1 works fine with 2023.1.6

CRogers:
changelog bot?

CRogers:
:+1:

crogoz:
👍

CRogers:
:+1:

felixdesouza:
👍

UpperLevel-zz:
I'm using version 2.46.0 and the problem still occurres. My intelliJ version is 2024.1.1. The buildnumber is #IU-241.15989.150. Anyone else has the same problem

PR #935, opened 2023-10-03
Java 21: Pattern Matching + StringTemplate JEP-430 + Unnamed JEP-445

npeters:
## Before this PR
May error on parsing Java 21 file.

## After this PR
Support of different JEP of Java 21 and preview
- JEP-430 StringTemplates
- JEP-440 Record Patterns
- JEP-441 Pattern Matching switch (with guard) 
- JEP-443 Unamed Pattern

### Build 
- I added a independent gradle module to build all Java 21 code but the gradle do not support to build groovy code with java 21: it should be first compile with java 17 and you can compile the project with java 21 😞
- I added a hack on palantir-java-format gradle script to copy all the java 21 class file on the jar. 😞
- Need to  add --enable-preview with Java 21.

### Code
- "StringTemplates" was really hard to support:  I did some change on `JavaInput.java` and `visitStringTemplate`. The complexity of this part is really to high.
- I copy and past "visitCase" from Java14 file and the guard param.

PR #1079, opened 2024-04-27
Gradle: Multi Compilation 21< and 21

npeters:
## Before this PR
This is the first step of my proposition, I did on #935
Formatting code is function of the jdk version but gradle configuration can not be build 

## After this PR
the package java/java21 is build with the jdk 21 and the other are build a other jdk

## Possible downsides?
.


npeters:
On CI Gradle javaToolChains did not found a Java 21

koppor:
Try the foojay toolchains plugin!

PR #1089, opened 2024-05-21
Stops optimizing imports via palantir-java-formatter

crogoz:
## Before this PR

Optimizing imports via palantir-java-format is slow. 
Previously (before switching to the new api: https://github.com/palantir/palantir-java-format/pull/1078), we would use Intelij's import optimizations (see: [old implementation](https://github.com/palantir/palantir-java-format/blob/2ec339a038e24c78b27509af9f2c7cd20ac798d5/idea-plugin/src/main/java/com/palantir/javaformat/intellij/CodeStyleManagerDecorator.java#L52) weren't overriding the [optimizeImports](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-api/src/com/intellij/psi/codeStyle/JavaCodeStyleManager.java#L65) implementation)

## After this PR

We stopped doing the import optimization via the `palantir-java-format`. Now, it gets delegated to Intelij.

==COMMIT_MSG==
Stops optimizing imports via palantir-java-formatter
==COMMIT_MSG==


idea-plugin/src/test/java/com/palantir/javaformat/intellij/PalantirJavaFormatImportOptimizerTest.java
```diff
@@ -95,6 +96,7 @@ public void tearDown() throws Exception {
     }
 
     @Test
+    @Disabled("Disabled the import optimization via palantir-java-format")
```
CRogers: I'd delete these tests entirely.

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -61,7 +61,7 @@ protected FormattingTask createFormattingTask(@NotNull AsyncFormattingRequest re
 
     @Override
     public @NotNull Set<Feature> getFeatures() {
-        return Set.of(Feature.FORMAT_FRAGMENTS, Feature.OPTIMIZE_IMPORTS);
+        return Set.of(Feature.FORMAT_FRAGMENTS);
```
CRogers, line 63: Should we return an empty set (or not override if it's set up like that) `getImportOptimizers`?

CRogers (approved):

PR #1115, opened 2024-08-13
Helpful Remediation in Known Error Situation

bmarcaur:
Helped someone debug this locally. Figured adding these "docs" inline might save someone grief in the future.

### DIFF
--- a/idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
+++ b/idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
@@ -139,13 +139,20 @@ final class FormatterProvider {
                 .map(Path::of)
                 .map(sdkHome -> sdkHome.resolve("bin").resolve("java" + (SystemInfo.isWindows ? ".exe" : "")))
                 .filter(Files::exists)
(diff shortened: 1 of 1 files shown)


idea-plugin/src/main/java/com/palantir/javaformat/intellij/FormatterProvider.java
```diff
@@ -139,13 +139,20 @@ private static Path getJdkPath(Project project) {
                 .map(Path::of)
                 .map(sdkHome -> sdkHome.resolve("bin").resolve("java" + (SystemInfo.isWindows ? ".exe" : "")))
                 .filter(Files::exists)
-                .orElseThrow(() -> new IllegalStateException("Could not determine jdk path for project " + project));
+                .orElseThrow(() ->
+                    new IllegalStateException("Could not determine JDK path for project: " + project.getName()));
```
bmarcaur: Printing the project directly creates error messages like so:

```
Could not determine jdk version for project Project(name=product, containerState=COMPONENT_CREATED, componentStore=/path/to/product)
```

Which feels slightly leaky and generally not that useful.

CRogers (approved):

PR #1127, opened 2024-09-11
Use com.palantir.external-publish-intellij for idea plugins

FinlayRJW:
## Before this PR
Had the same copy and pasted block of code for publishing intellij plugins which was not centrally managed

## After this PR
Migrated to the `com.palantir.external-publish-intellij` plugin to handle the publication of intellij plugins

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,11 @@ buildscript {
     }
 }
 
+plugins {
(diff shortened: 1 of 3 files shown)


FinlayRJW:
Need to update the version of `com.palantir.gradle.externalpublish:gradle-external-publish-plugin` before merging

CRogers (approved):

PR #1155, opened 2024-10-21
Cherry-pick upstream fixes for switch expressions with pattern matching

eholmer:
## After this PR
Cherry-pick the relevant commits to support java 21 style pattern matching in switch expressions.

Also setting java runtime to 21
==COMMIT_MSG==
Cherry-pick upstream changes to support java 21 style pattern matching
==COMMIT_MSG==


build.gradle
```diff
@@ -97,3 +95,8 @@ idea.project.ipr.withXml { xml ->
 jdks {
     daemonTarget = 17
 }
+
+javaVersions {
+    libraryTarget = 21
```
CRogers, line 100: I think one of the tricky parts here is IntelliJ runs with 17 rather than 21. In the IntelliJ plugin, we're probably going to have fork out to a new java process to run the formatting in a 21 JDK (there is already some code that does this iirc but it's slow. This is probably the impetus to get us to always fork but persist the JDK for a bit to make it fast).

CRogers, line 100: Will need to test how this works in gradle too. I suspect while we still support Java 17 in our gradle plugins we're going to have to do something a bit cleverer here too. This may also be the impetus to get us off spotless (or just upgrade every project to use 21 Gradle runtime).

fawind: This is the existing code that uses the configured project JDK for running the formatter: https://github.com/palantir/palantir-java-format/pull/562

But yeah, it's slow given it spins up a new JVM for each format command.

CRogers, line 100: And there is eclipse... which I have no idea about. Perhaps now is the moment to remove support - we do not support eclipse at all internally at Palantir any more.

CRogers: I guess we could make it do the same external JDK spinning up code we will end up doing for IntelliJ

crogoz:
Thanks @eholmer! I moved your commits & cherry-picks in https://github.com/palantir/palantir-java-format/pull/1211. 
I'll close this PR in favor of the new one.

PR #1180, opened 2024-12-07
Prefer in-process formatter when javac modules are exported

morokosi:
## Before this PR

- When using recent version of IntelliJ (shipped with JDK17+), the plugin uses bootstrapping formatter to avoid issues with internal API encapsulation
- Developers may suffer from slow formatting because the bootstrapping formatter is not as fast as the in-process formatter

## After this PR

- This PR allows optional use of in-process formatter when javac modules are exported in plugin vm process (specified in Custom VM Options)


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @morokosi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1188, opened 2025-01-09
[Enhancement] Only Send Updates When Formatting Changed

bmarcaur:
## Before this PR
Every format action caused the current buffer to be marked as modified.

## After this PR
This was due to us updating the text on every format even if it was unchanged. Now it only updates when something was formatted.

Way more context than you could ever want: https://youtrack.jetbrains.com/issue/IJPL-162382


idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -88,11 +88,21 @@ public void run() {
...
+
             try {
                 String formattedText = applyReplacements(
                         request.getDocumentText(),
                         formatterService.get().getFormatReplacements(request.getDocumentText(), toRanges(request)));
-                request.onTextReady(formattedText);
+
+                // The Javadoc of this API says that you should set it to null when the document is unchanged
+                // We should not be trying to format a document that is already formatted
+                if (preFormatText.equals(formattedText)) {
```
bmarcaur, line 114: Leveraging https://github.com/JetBrains/intellij-community/blob/9bcc90a5987ed7ac54336535e2dfa16eaddb0a3e/platform/code-style-api/src/com/intellij/formatting/service/AsyncFormattingRequest.java#L62 to dirty the buffer less

carterkozak: In this case, does `getFormatReplacements` return an empty list?

bmarcaur: uh, shoot I broke down my debug setup. Let me turn it back on

bmarcaur: Yeah you right..

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -89,10 +89,17 @@ public void run() {
...
-                        request.getDocumentText(),
-                        formatterService.get().getFormatReplacements(request.getDocumentText(), toRanges(request)));
-                request.onTextReady(formattedText);
+                List<Replacement> replacements =
+                        formatterService.get().getFormatReplacements(request.getDocumentText(), toRanges(request));
+
+                // The Javadoc of onTextReady API says that you should set it to null when the
+                // document is unchanged. But an even better version is to simply not attempt
+                // to format a document that is already formatted
+                if (replacements.isEmpty()) {
```
bmarcaur: Damn so it would turn out that even when the file is fully formatted it still produces a replacement which looks like:

image

A full file replacement with the same content... Seems like a bug in the formatter service.

CC @crogoz @carterkozak :(

bmarcaur: To avoid having to remote debug your intellij, here is a failing test that exemplifies the issue: https://github.com/palantir/palantir-java-format/pull/1188/files#diff-a6622bd2788dface780176f19bbaecce6308bb0a826e6e6721b0c713bc0e2c68R450

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java
```diff
@@ -101,18 +102,21 @@ public void run() {
...
-                            replacements.stream()
-                                    .map(Replacement::getReplaceRange)
-                                    .collect(Collectors.toSet())));
+
+                // Note: I attempted to implement this using getFormatReplacements and elide an update when
+                // there were no replacements. There is a bug in the underlying formatter where it _always_
+                // returns a change. Thus we must perform a content aware diff / branching.
+                String formattedText = applyReplacements(
+                        request.getDocumentText(),
+                        formatterService.get().getFormatReplacements(request.getDocumentText(), toRanges(request)));
```
crogoz: We should probably use the `preFormatText` instead:

```suggestion
                String formattedText = applyReplacements(
                        preFormatText,
                        formatterService.get().getFormatReplacements(preFormatText, toRanges(request)));
```


same in the in debug message above: https://github.com/palantir/palantir-java-format/pull/1188/files?new_files_changed=true#diff-65737b320acf0cb984bb55ee91a1aee8fd41a4b53051c53f277e652effb9f85bR102

palantir-java-format/src/test/java/com/palantir/javaformat/java/FormatterTest.java
```diff
@@ -452,4 +455,30 @@ void canParse_java9_private_interface_methods() {
...
+        String reformattedClass = formatter.formatSourceAndFixImports(formattedClass);
+
+        // The formatSourceAndFixImports code path flows through getFormatReplacements and
+        // this shows that despite the replacements happening they are superfluous
+        assertThat(formattedClass).isEqualTo(reformattedClass);
+
+        // Proof that the replacements are produced when fully formatted
+        assertThat(formatter.getFormatReplacements(
+                        formattedClass, List.of(Range.closedOpen(0, formattedClass.length()))))
+                .isEmpty();
```
crogoz: should we make the test successful instead ?

```suggestion
        // Proof that the replacements are produced when fully formatted
        List<Replacement> replacements =
                formatter.getFormatReplacements(formattedClass, List.of(Range.closedOpen(0, formattedClass.length())));
        assertThat(replacements.size()).isEqualTo(1);
        assertThat(replacements.getFirst().getReplaceRange()).isEqualTo(Range.closedOpen(0, formattedClass.length()));
```

bmarcaur: So I considered that, but its kinda weird. Because if it "regresses" that means the underlying bug is fixed... which now that I say it I could put a really clear error message that says revert this change and instead use the API properly.

crogoz (approved):

PR #1194, opened 2025-01-22
Initial Graal native-image creation & publishing with Benchmarks

crogoz:
## After this PR

- Creates the Graal native-image
- Publishing logic for the native-images
- Circle CI builds for linux (arm & amd) + macos (**_without_** Intel support: https://circleci.com/docs/using-macos/#available-resource-classes)
==COMMIT_MSG==
Graal native-image and Benchmarks
==COMMIT_MSG==

FLUPS: 
- Add windows build
- Use gradle-utils' Os & Arch classes: https://github.com/palantir/gradle-utils/pull/297
- Generate the native image metadata using the native-build plugin
- Once everyone uses the native-image, drop --add-exports GRADLE_OPTS https://github.com/palantir/palantir-java-format/pull/1194#discussion_r1949181301


build.gradle
```diff
@@ -98,4 +99,10 @@ idea.project.ipr.withXml { xml ->
 
 jdks {
     daemonTarget = 17
+
+    jdk(23) {
+        distribution = 'graalvm-ce'
+        jdkVersion = '23.0.1'
```
crogoz, line 105:
```
> Task :setupJdks

 + Options
     | Auto-detection:     Disabled
     | Auto-download:      Disabled

 + Amazon Corretto JDK 11.0.25+9-LTS
     | Location:           /Users/crogoz/.gradle/gradle-jdks/amazon-corretto-11.0.25.9.1
     | Language Version:   11
     | Vendor:             Amazon Corretto
     | Architecture:       aarch64
     | Is JDK:             true
     | Detected by:        Gradle property 'org.gradle.java.installations.paths'

 + Amazon Corretto JDK 17.0.13+11-LTS
     | Location:           /Users/crogoz/.gradle/gradle-jdks/amazon-corretto-17.0.13.11.1
     | Language Version:   17
     | Vendor:             Amazon Corretto
     | Architecture:       aarch64
     | Is JDK:             true
     | Detected by:        Current JVM

 + Amazon Corretto JDK 21.0.5+11-LTS
     | Location:           /Users/crogoz/.gradle/gradle-jdks/amazon-corretto-21.0.5.11.1
     | Language Version:   21
     | Vendor:             Amazon Corretto
     | Architecture:       aarch64
     | Is JDK:             true
     | Detected by:        Gradle property 'org.gradle.java.installations.paths'

 + GraalVM Community JDK 23.0.1+11-jvmci-b01
     | Location:           /Users/crogoz/.gradle/gradle-jdks/graalvm-ce-23.0.1
     | Language Version:   23
     | Vendor:             GraalVM Community
     | Architecture:       aarch64
     | Is JDK:             true
     | Detected by:        Gradle property 'org.gradle.java.installations.paths'


```

gradle/gradle-jdks-functions.sh
```diff
@@ -141,11 +141,11 @@ install_and_setup_jdks() {
       case "$distribution_url" in
         *.zip)
           distribution_name=${distribution_url##*/}
-          curl -C - "$distribution_url" -o "$distribution_name"
+          curl -L -C - "$distribution_url" -o "$distribution_name"
```
crogoz: just ran `./gradlew setupJdks` - same below

palantir-java-format/src/main/resources/META-INF/native-image/jni-config.json
```diff
@@ -0,0 +1,18 @@
+[
```
crogoz, line 1: these were generated via https://pl.ntr/2qm

palantir-java-format/build.gradle
```diff
@@ -30,6 +41,36 @@ dependencies {
...
+            javaLauncher = javaToolchains.launcherFor {
+                languageVersion = JavaLanguageVersion.of(23)
+                vendor = JvmVendorSpec.GRAAL_VM
+            }
+            imageName = 'palantir-java-format'
+            mainClass = mainClassName
+            fallback = false
+            sharedLibrary = false
+            configurationFileDirectories.from(file('src/main/resources/META-INF/native-image'))
+            buildArgs.addAll("-H:ConfigurationFileDirectories=${projectDir}/src/main/resources/META-INF/native-image",
```
crogoz, line 75: same args as in google-java-format: https://github.com/google/google-java-format/blob/eb0b7682f15ec6a2bd03d12b53bdf951dd924266/core/pom.xml#L282-L295

palantir-java-format-benchmarks/src/jmh/java/com/palantir/javaformat/BenchmarkMultiFiles.java
```diff
@@ -0,0 +1,110 @@
...
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+@Measurement(iterations = 2, time = 2)
+@Warmup(iterations = 2, time = 4)
+public class BenchmarkMultiFiles {
```
crogoz, line 43: Screenshot 2025-01-31 at 9 53 19 AM

.circleci/template.sh
CRogers, line 1: If we're moving off the template, I'd just delete this whole file.

.circleci/config.yml
```diff
@@ -78,6 +78,40 @@ jobs:
...
+    resource_class: macos.m1.medium.gen1
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.workers.max=2 -Dorg.gradle.jvmargs='-Xmx2147483648 --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -XX:MaxRAM=8g -XX:+CrashOnOutOfMemoryError
+    steps:
+      - checkout
+      - run: ./gradlew --parallel --stacktrace --continue --max-workers=2 nativeCompile -Porg.gradle.java.installations.fromEnv=JAVA_11_HOME,JAVA_17_HOME,JAVA_21_HOME,JAVA_HOME
+      - *check-no-files-changed
+      - persist_to_workspace:
+          root: /Users/distiller
```
CRogers, line 112: Is `distiller` the username circle sets by default?

.circleci/config.yml
```diff
@@ -78,6 +78,40 @@ jobs:
...
+  build-macos:
+    macos:
+      xcode: 15.4.0
+    resource_class: macos.m1.medium.gen1
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.workers.max=2 -Dorg.gradle.jvmargs='-Xmx2147483648 --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -XX:MaxRAM=8g -XX:+CrashOnOutOfMemoryError
+    steps:
+      - checkout
+      - run: ./gradlew --parallel --stacktrace --continue --max-workers=2 nativeCompile -Porg.gradle.java.installations.fromEnv=JAVA_11_HOME,JAVA_17_HOME,JAVA_21_HOME,JAVA_HOME
```
CRogers: We don't need the `fromEnv` stuff any more

.circleci/config.yml
```diff
@@ -78,6 +78,40 @@ jobs:
...
+          root: /home/circleci
+          paths: [ project/palantir-java-format/build/native/nativeCompile ]
+      - store_artifacts: { path: ~/artifacts }
+
+  build-macos:
+    macos:
+      xcode: 15.4.0
+    resource_class: macos.m1.medium.gen1
+    environment:
+      GRADLE_OPTS: -Dorg.gradle.workers.max=2 -Dorg.gradle.jvmargs='-Xmx2147483648 --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED'
```
CRogers, line 105: Actually, after we use Graal for the formatter I think we can remove this stuff from every repo! That will be good.

.circleci/config.yml
```diff
@@ -78,6 +78,40 @@ jobs:
...
+      _JAVA_OPTIONS: -XX:ActiveProcessorCount=4 -XX:MaxRAM=8g -XX:+CrashOnOutOfMemoryError
+    steps:
+      - checkout
+      - run: ./gradlew --parallel --stacktrace --continue --max-workers=2 nativeCompile -Porg.gradle.java.installations.fromEnv=JAVA_11_HOME,JAVA_17_HOME,JAVA_21_HOME,JAVA_HOME
+      - *check-no-files-changed
+      - persist_to_workspace:
+          root: /Users/distiller
+          paths: [ project/palantir-java-format/build/native/nativeCompile ]
```
CRogers, line 113: These paths are all the same - is the OS/arch in the filename?

palantir-java-format/src/main/resources/META-INF/native-image/jni-config.json
CRogers: Ideally we'd generate these automatically somehow as part of the build (or verify the checked in versions are correct) - if not, can we at least make a gradle task (or even just some instructions) to regenerate them that are not hidden via that link?

palantir-java-format-benchmarks/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    jmhImplementation 'org.assertj:assertj-core'
+    jmhImplementation 'org.junit.jupiter:junit-jupiter'
+    jmhImplementation 'org.openjdk.jmh:jmh-core'
+    jmhImplementation 'org.openjdk.jmh:jmh-generator-annprocess'
+}
+
+tasks.named('processJmhResources', ProcessResources.class) {
+    from(project(':palantir-java-format').tasks.named('fatJar')) {
+        rename { 'palantir-java-format-all.jar' }
+    }
```
CRogers: Referring to tasks in other gradle projects is not good and Gradle will eventually ban when Isolated Projects come. Really we should [use outgoing configurations](https://docs.gradle.org/current/userguide/cross_project_publications.html).

Rather than making a fat jar, I think for this one you can just:
1. Make a new configuration in this project that has `project(':palantir-java-format')` as a dependency.
2. Make a `Copy` task (perhaps even a `Sync` task) that copies the configuration to a directory.
3. [Add a environment variable](https://github.com/melix/jmh-gradle-plugin/blob/493a4f0ca583d95ee5a61f20ae840d8d83eaa108/src/main/java/me/champeau/jmh/JmhParameters.java#L150) to the jmh task that passes in the output dir of the copy task (this would be easier in java, we could even forgo the Copy task and just do `configuration.map(config -> config.getFiles())` perhaps...).

palantir-java-format-benchmarks/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+    from(project(':palantir-java-format').tasks.named('nativeCompile')) {
+        rename { 'palantir-java-format'}
+    }
```
CRogers: We should make an outgoing artifact/configuration for nativeImage. I think you just need to do something like (wildly untested):

```gradle
configurations {
    nativeImage {
        outgoing.artifact(tasks.named('nativeImage'))
    }
}
```

CRogers: Then you make a configuration in this project and add a dep like `consumingProjectConfiguration project(name: ':palantir-java-format', configuration: 'nativeImage')`, then use that

.circleci/config.yml
crogoz: Yes, but I couldn't find any documentation regarding this setup - besides some issues is Circle.

.circleci/config.yml
crogoz: Yes, the native images will be of the format: `palantir-java-format-<os>-<arch>`

palantir-java-format/src/main/resources/META-INF/native-image/jni-config.json
crogoz: There is an option to collect the metadata automatically: https://graalvm.github.io/native-build-tools/latest/end-to-end-gradle-guide.html#collect-metadata but I am yet to figure out how to pass arguments to the native-image during the collection of the metadata. I was thinking of doing this as part of a FLUP PR.

palantir-java-format-benchmarks/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+import me.champeau.jmh.JMHTask
+
+tasks.named('jmh', JMHTask.class) {
+    inputs.files(configurations.consumingNativeImage)
+    inputs.files(configurations.consumingPalantirJars)
+    environment.put('NATIVE_IMAGE_CLASSPATH', configurations.consumingNativeImage.asPath)
+    environment.put('JARS_CLASSPATH', configurations.consumingPalantirJars.collect { it.absolutePath }.join(":"))
```
CRogers: to make this both fully lazy and automatically wire up the inputs, it needs to be `configurations.named('consumingPalantirJars').map { conf -> conf.collect { it.absolutePath }.join(":") }`. Otherwise this configuration gets resolved when the `jmh` task is realised, which could be whenever.

This should also allow us to get rid of the `inputs.files` above.

palantir-java-format-benchmarks/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+}
+
+import me.champeau.jmh.JMHTask
+
+tasks.named('jmh', JMHTask.class) {
+    inputs.files(configurations.consumingNativeImage)
+    inputs.files(configurations.consumingPalantirJars)
+    environment.put('NATIVE_IMAGE_CLASSPATH', configurations.consumingNativeImage.asPath)
```
CRogers: Similarly for this, needs to be `configurations.named('consumingNativeImage').map { it.asPath }`

palantir-java-format-benchmarks/build.gradle
```diff
@@ -0,0 +1,36 @@
+apply plugin: 'distribution'
+apply plugin: 'me.champeau.jmh'
+
+configurations {
+    consumingNativeImage
+    consumingPalantirJars
```
CRogers: Not sure `consuming`/`consumable` add much here - I'd just call them something like `formatterNativeImage` and `formatterNativeJars`.

palantir-java-format/build.gradle
```diff
@@ -89,3 +139,70 @@ tasks.named("jar", Jar) {
...
+        return "aarch64"
+    }
+    if (Set.of("x86", "i686").contains(osArch)) {
+        return "x86"
+    }
+    throw new GradleException(String.format("Invalid Operating System %s", osArch))
+}
+
+configurations {
+    consumableNativeImage {
```
CRogers: I'd just call this `nativeImage` perhaps.

palantir-java-format-benchmarks/build.gradle
crogoz: I still needed to add the `inputs.files` with this lazy configuration, otherwise the `nativeCompile` task wouldn't be run before `jmh` https://github.com/palantir/palantir-java-format/pull/1194/commits/760922c401bd7cdc6b244d3b4f75b59561ce90a3

CRogers (approved):
:+1: Let's hope this publishing works 🤞🏻

PR #1200, opened 2025-01-30
Feature flagged native image palantir-java-format

crogoz:
## After this PR
Adds support for running the native-image based formatter inside gradle and Intellij. 
The enablement of the palantir-java-format native image formatter is based on: 
- OS & arch compatibility. Atm we are only supporting linux & macos M1 native-images - see: https://github.com/palantir/palantir-java-format/pull/1194/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R160
- Setting the `palantir.native.formatter=true` Gradle property

Updates the `.idea/palantir-java-format.xml` settings file with the classpath to get the native image. Adds a flag to specify if the native image should be used.

The palantir-java-format settings panel will show if we are using the native-image or the legacy palantir formatter

==COMMIT_MSG==
Feature flagged native image palantir-java-format
==COMMIT_MSG==


gradle-palantir-java-format/build.gradle
```diff
@@ -88,3 +88,7 @@ def writeImplClasspath = tasks.register("writeImplClasspath") {
 tasks.named("test").configure {
     dependsOn(writeImplClasspath)
 }
+
+tasks.withType(ProcessResources.class).configureEach {
+    from(project(':palantir-java-format').tasks.named('nativeCompile')) {}
+}
```
CRogers: We'll want to download them separately using a configuration right? Rather than just shoving them in the jar and making it enormous

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
```diff
@@ -18,36 +18,87 @@
...
-                        .getDependencies()
-                        .create(ImmutableMap.of(
-                                "group", "com.palantir.javaformat",
-                                "name", "palantir-java-format",
-                                "version",
-                                        JavaFormatExtension.class.getPackage().getImplementationVersion())));
-            });
+        String implementationVersion = JavaFormatExtension.class.getPackage().getImplementationVersion();
+
+        Configuration configuration = rootProject.getConfigurations().create(CONFIGURATION_NAME);
```
CRogers: `.register` and put the method calls inside the register action

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
```diff
@@ -18,36 +18,87 @@
...
+                            "group",
+                            "com.palantir.javaformat",
+                            "name",
+                            "palantir-java-format-native",
+                            "version",
+                            implementationVersion,
+                            "classifier",
+                            String.format("nativeImage-%s", getCurrentOs()),
+                            "ext",
+                            "exe")));
```
CRogers: This is pretty hard to read - I'd either use the string version `String.format("com.palantir.javaformat:palantir-java-format-native:%s:%s@%s")` or at least `builder()` variant for the map so each part lines up

CRogers: shouldn't have an exe extension for non-windows binaries

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
```diff
@@ -18,36 +18,87 @@
...
+        });
+        nativeConfiguration
+                .getAttributes()
+                .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "executable-nativeImage");
+        rootProject.getDependencies().registerTransform(ExecutableTransform.class, transformSpec -> {
+            transformSpec.getFrom().attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "exe");
```
CRogers: How is this working? Does Gradle set this attribute to be the extension of the dependency?

gradle-palantir-java-format/build.gradle
crogoz: Yes, I needed to include the native image in the resources for some tests. I will push a change which will only affect the test resources.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
crogoz: Unfortunately I need an extension, if I leave the extension empty, the file will be considered as a jar. 
Since on unix I don't think the extension matters, I left it with `exe` here as well - I can change it to something else - but had troubles finding a name for the extension 😅

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
crogoz: Yes, from my testing - I saw it mentioned in a few other places in their codebase: https://github.com/gradle/gradle/blob/2d2ed60fc306a1101536500489c243f2d240af68/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/attributes/immutable/artifact/ImmutableArtifactTypeRegistry.java#L118

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatProviderPlugin.java
CRogers: Mmm, I guess we could make it `.bin` for macos/linux. It's not ideal, but I think it's going to confuse any external people using it having it be `.exe`.

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatConfigurable.java
```diff
@@ -115,6 +113,16 @@ public boolean isModified() {
     @Override
     public void disposeUIResources() {}
 
+    private static String isUsingNativeImage(PalantirJavaFormatSettings settings) {
```
crogoz, line 116: Screenshot 2025-02-18 at 2 50 31 PM

gradle-palantir-java-format/build.gradle
```diff
@@ -74,15 +74,38 @@ configurations {
...
+        String.format("%s.bin", fileName)
+    }
+    into("$buildDir/nativeImage")
 }
 
 def writeImplClasspath = tasks.register("writeImplClasspath") {
+    inputs.files(tasks.named("copyNativeImage"))
     doLast {
         file("$buildDir/impl.classpath").text = configurations.impl.asPath
+        file("$buildDir/nativeImage.path").text = tasks.named("copyNativeImage").get().getOutputs().getFiles().getSingleFile().listFiles()[0].getAbsolutePath()
```
crogoz, line 108: required by: https://github.com/palantir/palantir-java-format/pull/1200/files#diff-730f59a46b5a223a716041a78825ed9f9a88c8dbef24bd8c5a07fb2a72adcd22R36. The native image file needs to specify ".bin" extension (this is the artifactType we expect for palantirjavaFormatNative configuration)

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatPlugin.java
```diff
@@ -16,42 +16,77 @@
...
         project.getPlugins().apply(PalantirJavaFormatSpotlessPlugin.class);
 
         project.getPlugins().withId("java", p -> {
-            project.getTasks().register("formatDiff", FormatDiffTask.class);
 
             // TODO(dfox): in the future we may want to offer a simple 'format' task so people don't need to use
             // spotless to try out our formatter
+            project.getTasks().register("formatDiff", FormatDiffTask.class, task -> {
+                if (NativeImageFormatProviderPlugin.isNativeImageSupported()
+                        && NativeImageFormatProviderPlugin.shouldUseNativeImage(project)) {
```
CRogers: Perhaps we should just make `shouldUseNativeImage` include `isNativeImageSupported`. I don't think we're ever going to use the native image if it's not supported?

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatPlugin.java
```diff
@@ -16,42 +16,77 @@
...
+                .map(conf -> conf.getSingleFile().toPath()));
+    }
+
+    public abstract static class FormatDiffTask extends DefaultTask {
+
+        private static Logger log = Logging.getLogger(FormatDiffTask.class);
+
+        @org.gradle.api.tasks.Optional
+        @Input
+        abstract Property<Path> getNativeImage();
```
CRogers: This should really be an `@InputFile` - can it be a `@Optional @InputFile RegularFileProperty`? If there's problems with the optional, you can do the trick of having a `ConfigurableFileCollection` that has one or more files.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/ExecutableTransform.java
```diff
@@ -0,0 +1,73 @@
...
+            Set<PosixFilePermission> existingPermissions = Files.getPosixFilePermissions(pathToExe);
+            Files.setPosixFilePermissions(
+                    pathToExe,
+                    Stream.concat(
+                                    existingPermissions.stream(),
+                                    Stream.of(
+                                            PosixFilePermission.OWNER_EXECUTE,
+                                            PosixFilePermission.GROUP_EXECUTE,
+                                            PosixFilePermission.OTHERS_EXECUTE))
+                            .collect(Collectors.toSet()));
```
CRogers, line 72: I'm guessing on windows, we just don't run this transform at all (otherwise it would fail?)?

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final Logger log = Logging.getLogger(NativeImageFormatProviderPlugin.class);
+    static final String NATIVE_CONFIGURATION_NAME = "palantirJavaFormatNative";
+
+    @Override
+    public void apply(Project rootProject) {
+        Preconditions.checkState(
+                rootProject == rootProject.getRootProject(),
+                "May only apply com.palantir.java-format-provider to the root project");
+
+        if (!isNativeImageSupported()) {
```
CRogers: Should this check `shouldUseNativeImage()`?

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+            return;
+        }
+        String implementationVersion = JavaFormatExtension.class.getPackage().getImplementationVersion();
+        OperatingSystem operatingSystem = OperatingSystem.get();
+        rootProject.getConfigurations().register(NATIVE_CONFIGURATION_NAME, conf -> {
+            conf.setDescription("Internal configuration for resolving the palantir-java-format native image");
+            conf.setVisible(false);
+            conf.setCanBeConsumed(false);
+            conf.setCanBeResolved(true);
+            conf.defaultDependencies(deps -> {
```
CRogers, line 51: I think this use of `defaultDependencies` will mean if someone else adds a dep these will not get added; not sure if that's what we want.

CRogers: Oh, this is what makes the tests work by adding a `file(...)` dep to this configuration?

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/ExecutableTransform.java
CRogers: Ah yes, we don't register the transform at all.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -0,0 +1,89 @@
...
+                    .getFrom()
+                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem));
+            transformSpec.getTo().attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "executable-nativeImage");
+        });
+    }
+
+    public static boolean isNativeImageSupported() {
+        OperatingSystem os = OperatingSystem.get();
+        return os.equals(OperatingSystem.LINUX_GLIBC)
+                || (os.equals(OperatingSystem.MACOS) && Architecture.get().equals(Architecture.AARCH64));
```
CRogers, line 78: I think it needs to be `aarch64 && (linux_glibc || macos)`, we don't support linux_glibc aarch64 either.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/ExecutableTransform.java
```diff
@@ -0,0 +1,73 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to set execute permissions on native-image", e);
+        }
+    }
+
+    interface Parameters extends TransformParameters {}
```
CRogers: Unused I think

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/ExecutableTransform.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private static Logger logger = Logging.getLogger(ExecutableTransform.class);
+
+    @InputArtifact
+    public abstract Provider<FileSystemLocation> getInputArtifact();
+
+    @Override
+    public void transform(TransformOutputs outputs) {
+        File inputFile = getInputArtifact().get().getAsFile();
+        makeFileExecutable(inputFile.toPath());
```
CRogers: This is modifying the executable in the cache, no? Really we should copy it first to not modify it.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
crogoz: we support both aarch64 & amd64 for linux: https://github.com/palantir/palantir-java-format/blob/develop/.circleci/config.yml#L22; and https://repo.maven.apache.org/maven2/com/palantir/javaformat/palantir-java-format-native/2.52.0/

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/ExecutableTransform.java
CRogers: This has other strange effects - like when I run with a `publishToMavenLocal`d executable, it changes the file in the maven local `~/.m2` repo

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
crogoz: > Oh, this is what makes the tests work by adding a file(...) dep to this configuration?
yes, exactly

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java
```diff
@@ -0,0 +1,72 @@
...
+
+        State(FileSignature pathToExe) {
+            this.pathToExe = pathToExe;
+        }
+
+        String format(ProcessRunner runner, String input) throws IOException, InterruptedException {
+            List<String> argumentsWithPathToExe =
+                    List.of(pathToExe.getOnlyFile().getAbsolutePath(), "--palantir", "-");
+            return runner.exec(input.getBytes(StandardCharsets.UTF_8), argumentsWithPathToExe)
+                    .assertExitZero(StandardCharsets.UTF_8);
```
CRogers, line 64: While using the native formatter in IntelliJ is noticeably snappier, we have the old problem in IntelliJ here now:

In spotless, we now start a process per file, where as the old one just run java code over and over that had already been classloaded. On one internal project format goes from taking [32 secs](https://pl.ntr/2qT) to [1min 20sec](https://pl.ntr/2qS).

This is quite a lot slower. Either it's because of starting all the processes or because the graaled code itself is slower (or both)

idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatSettings.java
```diff
@@ -91,6 +91,20 @@ Optional<List<URI>> getImplementationClassPath() {
...
+     * The path to the formatter nativeImage.
+     */
+    Optional<URI> getNativeImageClassPath() {
+        return state.nativeImageClassPath;
+    }
+
+    /**
+     * Whether to use the legacy implementation(getImplementationClassPath) or the new implementation(getNativeImageClassPath)
+     */
+    Boolean getUseNativeImageClassPath() {
```
CRogers: Does this handle the case where this config is not here at all? I think it's probably best to support the `useNativeImageClassPath` not existing for two reasons:

1. People could update the plugin then it might crash if the user opens the settings page on a project that hasn't resynced gradle yet and so doesn't have the config - I think I saw this yesterday.
2. If we ever want to remove it later, should make it optional so when people upgrade the gradle plugin before the intellij plugin it still works.

CRogers: Actually, can we just use whether or not `getNativeImageClassPath` exists to determine whether or not to use the native formatter? And just remove this boolean entirely? There's an invalid state possible state here atm (`useNativeImageClassPath = true`, `getNativeImageClassPath` does not exist) that we could remove.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java
CRogers: Discussed internally; spotless is incremental and build cached, so the damage is mitigated somewhat, especially for PR builds (both local and CI) that don't change the gradle build script (so don't invalidate the build cache). We can build the process management bit later. The speedup inside intellij from this change when modifying lots of files is immense.

CRogers (approved):

CRogers:
:+1:

PR #1208, opened 2025-02-12
Separate palantir-java-format-native project & fix publishing

crogoz:
## After this PR
Follow-up for https://github.com/palantir/palantir-java-format/pull/1194 that fixes the native image publishing. 
- Creates a new project `palantir-java-format-native` that builds the native image & deals with the publishing using `external-publish-custom` plugin

==COMMIT_MSG==
Separate palantir-java-format-native project & fix publishing
==COMMIT_MSG==


palantir-java-format-native/build.gradle
```diff
@@ -0,0 +1,111 @@
...
+
+// The image requires parameters to be passed to; eg. gw nativeRun -Pargs=--palantir,-i,<file>
+tasks.named("nativeRun", org.graalvm.buildtools.gradle.tasks.NativeRunTask) {
+    runtimeArgs.addAll(project.hasProperty('appArgs') ? project.appArgs.split(',') : [])
+}
+
+def osExtension = operatingSystem == OperatingSystem.MACOS ? 'bin' : 'exe'
+
+externalPublishing {
+    publishing {
```
CRogers: I don't think this is calling anything under `externalPublishing`. There is a single method [`publication`](https://github.com/palantir/gradle-external-publish-plugin/blob/8fe67d8680eb57e37442b417a8177a2a9e1658f2/src/main/java/com/palantir/gradle/externalpublish/ExternalPublishCustomExtension.java#L32) that needs to be called with a name and a closure to configure the maven publication. This is currently calling the `publications` extension that lives on `project` from the `maven-publish` plugin and the external publish plugin does not know to actually publish this one.

palantir-java-format-native/build.gradle
```diff
@@ -0,0 +1,111 @@
...
+        publications {
+            nativeImage(MavenPublication) {
+                fileTree("build/native/nativeCompile", { it.include("**/*") }).forEach { file ->
+                    artifact(file) {
+                        classifier = String.format("nativeImage-%s", getOsArchName(file))
+                        extension = osExtension
+                    }
+                }
```
CRogers: This feels odd with it not being lazy. If the tasks have not run yet when this is evaluated then there will be no artifacts for this publication. I guess it works because we attach workspaces before gradle is run for publish. But does this mean `publishToMavenLocal` run on a laptop doesn't work?

palantir-java-format-native/build.gradle
```diff
@@ -0,0 +1,111 @@
...
+def osExtension = operatingSystem == OperatingSystem.MACOS ? 'bin' : 'exe'
+
+externalPublishing {
+    publishing {
+        publications {
+            nativeImage(MavenPublication) {
+                fileTree("build/native/nativeCompile", { it.include("**/*") }).forEach { file ->
+                    artifact(file) {
+                        classifier = String.format("nativeImage-%s", getOsArchName(file))
+                        extension = osExtension
```
CRogers: Shouldn't this vary per artifact? Currently it seems to be fixed based on the OS running the publish job?

palantir-java-format-native/build.gradle
crogoz: I changed this in: https://github.com/palantir/palantir-java-format/pull/1208/commits/8456625f22b51b7930eb83adf6fb2ea640f5a5ab#diff-012bca426d515b4e67fdc32e660cf6ade9abd1d720f3f923e6b16101a72dc3c3R79

palantir-java-format-native/build.gradle
CRogers: `artifact` can apparently take anything that `Project#file` can, so we can give it providers or other tasks - I'm just not sure if it will try to rerun the tasks or not on the publish task (which won't work). If publishToMavenLocal doesn't currently work, maybe we need a separate publication not registered with external-publish-plugin that actually takes the tasks as inputs to the artifacts (for just publishToMavenLocal to work? We're going to want to be able to use that to test locally)?

crogoz: ~~It works locally once I added: https://github.com/palantir/palantir-java-format/pull/1208/commits/8456625f22b51b7930eb83adf6fb2ea640f5a5ab#diff-012bca426d515b4e67fdc32e660cf6ade9abd1d720f3f923e6b16101a72dc3c3R84~~

crogoz: ended up doing 2 separate publication depending if we are in CI or not - https://github.com/palantir/palantir-java-format/pull/1208/files#diff-012bca426d515b4e67fdc32e660cf6ade9abd1d720f3f923e6b16101a72dc3c3R77

CRogers (approved):

PR #1211, opened 2025-02-18
Java 21 with Graal support

crogoz:
## Before this PR

No java 21 support

## After this PR

Java 21 support that can run either through the native-image - by setting the gradle property `palantir.native.formatter=true` or - by default - using the java-based implementation.

Adds the following support: 
- JEP-441 Pattern Matching switch (with guard)
- JEP-443 Unnamed Patterns and Variables
- Fixes the comments indentation inside switch statements 

==COMMIT_MSG==
Java 21 Support with Graal & Java-based Implementations
==COMMIT_MSG==

FLUPs: 
- support reflowing strings in intellij see: https://github.com/palantir/palantir-java-format/pull/1211#discussion_r1987034457


palantir-java-format-21/src/test/resources/com/palantir/javaformat/java/StringWrapperIntegrationTest/0.input
CRogers, line 1: Moving all these tests and test files is pretty annoying. It feels like it will happen again in a future java version with source changes.

We also have the same problem with the `palantir-java-format-21` dependency in the Gradle plugin. When the next java version appears with the source changes, this will be changed too. I think we might want to decouple the gradle plugin, formatter and idea plugin at some point (to avoid making so many unnecessary releases, especially for the idea plugin), which will make dependency changes like this harder.

 I wonder if we can avoid this somehow - I think ideally we'd have a `palantir-java-format` that is empty (except for tests) which we just make depend on the latest (eg `palantir-java-format-21`). When the next version comes around we just change the dep to eg `palantir-java-format-25` but all the end users keep using `palantir-java-format`. The only downside to this is we'd need to move all the code in `palantir-java-format` once to `palantir-java-format-14` (or whatever it would be) which would lose all the history - maybe we could make a `palantir-java-format-all` instead?

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/SwitchRecord.output
```diff
@@ -34,10 +34,8 @@ record SwitchRecord(int i) {
                     int seven,
                     int eight,
                     int nine) ->
-                System.err.println("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor"
-                        + " incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis"
-                        + " nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo"
-                        + " consequat.");
+                System.err.println(
```
crogoz, line 37: there is a bug in palantir-java-format that doesn't wrap long strings when running through the Intellij plugin - https://github.com/palantir/palantir-java-format/blob/develop/palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java#L51 vs https://github.com/palantir/palantir-java-format/blob/develop/palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java#L70. I will fix this in a separate PR

palantir-java-format-spi/build.gradle
```diff
@@ -11,8 +11,3 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'com.fasterxml.jackson.core:jackson-databind'
 }
-
-tasks.withType(JavaCompile).named('compileJava') {
```
crogoz: moved all the *Compatibility blocks to the root build.gradle in the `subprojects` block. We need to set the sourceCompatibility for groovy sourceSets as well, so I think it is better to always set the source & target compatibility to 11

palantir-java-format/src/main/resources/META-INF/native-image/reflect-config.json
```diff
@@ -1,56 +0,0 @@
-[
-{
-  "name":"[Lcom.fasterxml.jackson.databind.deser.Deserializers;"
```
crogoz, line 3: old file that was generated with graal 21 - now we are using reachability-metadata: https://github.com/palantir/palantir-java-format/pull/1211/files#diff-08891f153c2557b2f9c4131cb3f735dda8476f2f55ed91e9227b769ebe069991R87

build.gradle
```diff
@@ -67,11 +71,12 @@ subprojects {
...
 
 javaVersions {
-    libraryTarget = 17
+    libraryTarget = 21
+    runtime = 21
 }
 
 jdks {
-    daemonTarget = 17
+    daemonTarget = 21
```
CRogers: Do we need daemonTarget 21?

crogoz: Not really, but since everything else would use java 21, I thought it is OK to move the daemon as well. 
Moved it back to 17 https://github.com/palantir/palantir-java-format/pull/1211/commits/5ca751c0dcc9baee080c7a2d4ca7614338cf72da

palantir-java-format-spi/build.gradle
CRogers: Why do we need to set the sourceCompatibility for the groovy source sets too? If we can, I'd like to contain this strangeness to just the place that needs, ie the code for running the actual formatter

CRogers: I think we might avoid being able to do this by keeping `libraryTarget = 17` on the root project (so the gradle plugin, idea plugin etc use 17) then putting `libraryTarget = 21` on the `palantir-java-format` and `palantir-java-format-spi` projects

build.gradle
```diff
@@ -67,7 +71,8 @@ subprojects {
 }
 
 javaVersions {
-    libraryTarget = 17
+    libraryTarget = 21
+    runtime = 21
```
CRogers, line 72: `libraryTarget` implies `runtime`, I think we can just leave this out to keep it like our other Gradle plugins.

palantir-java-format-spi/build.gradle
CRogers: Oops, I mean `target = 21` in the subprojects.

crogoz: changed in https://github.com/palantir/palantir-java-format/pull/1211/commits/ff5418204c120ea04c3f53023bf65f97a4f7478d

CRogers (approved):

crogoz:
👍

PR #1216, opened 2025-02-20
Fix MissingResourceRegistration for native-image 

crogoz:
## Before this PR

Fixes failures: 
```
Stderr:
Exception in thread "main" com.oracle.svm.core.jdk.resources.MissingResourceRegistrationError: The program tried to access the resource at path

   com/sun/tools/javac/resources/compiler_en.properties

without it being registered as reachable. Add it to the resource metadata to solve this problem. See https://www.graalvm.org/latest/reference-manual/native-image/metadata/#resources-and-resource-bundles for help
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.resources.MissingResourceRegistrationUtils.missingResource(MissingResourceRegistrationUtils.java:4
```	

In https://github.com/palantir/palantir-java-format/pull/1212, we introduced the flag `--exact-reachability-metadata` which would fail during build time if the native image is missing some resources.

## After this PR

I've regenerated the reachability-metadata using GraalVM 23. It was previously generated using GraalVM 21. 
==COMMIT_MSG==
Fix MissingResourceRegistration for native-image 
==COMMIT_MSG==


gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/ConfigureJavaFormatterXml.groovy
```diff
@@ -30,9 +30,11 @@ class ConfigureJavaFormatterXml {
             listItems.appendNode('option', [value: uri])
         }
         // configure nativeImageClasspath
-        nativeImageUri.ifPresent { URI uri ->
+        nativeImageUri.ifPresentOrElse({ URI uri ->
```
crogoz: if we remove the gradle property `palantir.native.formatter`, intellij needs to switch to using the bootstrapping jdk. therefore, we need to delete the nativeImageClassPath entry in the `palantir-java-format.xml` file. Fixes issue introduced in: https://github.com/palantir/palantir-java-format/pull/1200/commits/1f084a1592bc9b2ebbaf1ffcb3b26d73bb6aec33#diff-a2d2407a6a459747eaf9b40ec418e6ac7824c159c005b1f173623906db71a336L75

crogoz: moved to a different PR: https://github.com/palantir/palantir-java-format/pull/1217

carterkozak:
Is there any way we could have triggered the failure in tests?

crogoz:
> Is there any way we could have triggered the failure in tests?

This actually triggered the test failures: https://app.circleci.com/pipelines/github/palantir/palantir-java-format/3816/workflows/83ae9ee3-02bc-434f-a886-a64f721c1d5f/jobs/16378/tests but I think in this case the order of the commits was a little bit unfortunate, the PR that had the tests didn't merge with develop (that had the flag `--exact-reachability-metadata` set). So I only noticed it on develop once my PR got merged

carterkozak (approved):
Awesome,  that makes sense. Thanks!

PR #1226, opened 2025-03-07
Use modern infra to support multiple JDK versions

CRogers:
## Before this PR
This repo uses a lot of custom tooling rather than our "modern" tooling to set up and configure JDKs.

It does do something nifty though: to support the formatter being run on multiple different different JDK versions, it uses a newer compiler but keeps the target compatibility low. This means that the same jar works whichever JDK you run on it.

## After this PR
We keep doing the trick of outputting Java 11 class files - but now we use basline-java-versions (now possible after https://github.com/palantir/gradle-baseline/pull/3060).

I also removed all the other custom intellij settings customisation in this repo

==COMMIT_MSG==
Use modern infra to support multiple JDK versions
==COMMIT_MSG==


build.gradle
```diff
@@ -67,33 +66,9 @@ subprojects {
     }
 }
 
-/**
- * There is a bunch of truly snowflaky stuff going on this project, because we have this Java14InputAstVisitor class
- * which needs to implement some interface methods defined in com.sun.source.util.TreePathScanner, which varies
- * depending on what JVM the formatter is running on!
- */
-idea.project.ipr.withXml { xml ->
```
CRogers, line 75: This just doesn't seem to be necessary any more.

idea-plugin/build.gradle
```diff
@@ -69,14 +69,6 @@ dependencies {
     testImplementation 'org.junit.platform:junit-platform-launcher'
 }
 
-tasks.withType(JavaCompile).configureEach {
-    options.errorprone.disable 'StrictUnusedVariable'
-    // opt out of safe-logging checks which don't map correctly to idea plugins
-    options.errorprone.disable 'PreferSafeLogger'
-    options.errorprone.disable 'PreferSafeLoggableExceptions'
-    options.errorprone.disable 'PreferSafeLoggingPreconditions'
-}
```
CRogers, line 78: Now configured at the root level on all projects

palantir-java-format/build.gradle
```diff
@@ -30,40 +30,28 @@ dependencies {
...
+            'jdk.compiler/com.sun.tools.javac.parser',
+            'jdk.compiler/com.sun.tools.javac.tree',
+            'jdk.compiler/com.sun.tools.javac.util',
+            'jdk.compiler/com.sun.tools.javac.code',
+            'jdk.compiler/com.sun.tools.javac.api')
 }
 
-tasks.withType(Test).configureEach {
-    jvmArgs = jvmArgList
+tasks.withType(JavaCompile).named('compileJava') {
```
CRogers, line 44: By doing this just for `compileJava` we can use higher level language features in the tests

palantir-java-format/build.gradle
```diff
@@ -30,40 +30,28 @@ dependencies {
...
 }
 
 tasks.withType(Javadoc).configureEach {
-    // Allow access to internal javac apis
-    options.optionFiles << file('../gradle/javadoc.options')
-
-    if (JavaVersion.current() < JavaVersion.VERSION_14) {
-        exclude '**/Java14InputAstVisitor.java'
-    }
+    enabled = false
```
CRogers: We rarely if ever use javadoc - I'll need to check this doesn't break publishing

CRogers: I've just remembered that Maven Central requires javadoc jars - I'll try to get it to work

CRogers:
Actually going to do a release so we can check publishing still works/this works on lots of repos (I've tested manually on a couple)

crogoz (approved):

PR #1243, opened 2025-03-19
SpotlessApply uses by default the Java formatter

crogoz:
## Before this PR

The throughput for formatting big repos is decreased by using the native image (CE edition) - see internal metrics doc [here](https://pl.ntr/2sa) and [comment here](https://github.com/palantir/palantir-java-format/pull/1200#discussion_r1960098782)
java-based implementation [32 secs](https://pl.ntr/2qT) vs native-image implementation [1min 20sec](https://pl.ntr/2qS).

Even though spotlessApply is incremental and cacheable, in case the task cannot run incrementally, the performance is not good. 

## After this PR

When running `spotlessApply` tasks only  (to format full projects), we will switch to using: 
- Java-based implementation if the daemonTarget already is set to Java 21
- otherwise we will use the native-image until the rollout of Java 21 is finished everywhere.

formatDiff/Intellij will still use the native image formatter by default - the perf for running on a few files is better with native-image vs Java-based implementations.

==COMMIT_MSG==
SpotlessApply uses by default the Java formatter
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,25 @@ shortcut.
 
 ![Install plugin from disk](./docs/images/install_plugin_from_disk.png)
 
+## Java 21 Support
+
+In [1211](https://github.com/palantir/palantir-java-format/pull/1211) we shipped Java 21 support. In order to use the 
+Java 21 formatting capabilities, ensure that either: 
+
+- the Gradle daemon and the Intellij Project SDK are set to Java 21
+- or that the gradle property `palantir.native.formatter=true`. This will run the formatter as a native image, 
+- independent of the Gradle daemon/Intellij project JDK version. 
+
+### Native image formatter
+
+[This comment](https://github.com/palantir/palantir-java-format/issues/952#issuecomment-2575750610) explains why we 
+switched to a native image for the formatter. The startup time for the native image esp. in Intellij is >10x faster than
+spinning up a new process that does the formatting.
+However, the throughput of running the native image for a large set of files (eg. running `./gradlew spotlessApply`) is 
+considerably slower (eg. 30ms using the Java implementation vs 1m20s using the native image implementation). Therefore,
+when running the formatter from `spotlessApply` we will default to using the Java implementation (if the Java version >= 21).
+
+
 ## Future works
 
 - [ ] preserve [NON-NLS markers][] - these are comments that are used when implementing NLS internationalisation, and need to stay on the same line with the strings they come after.
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ Java 21 formatting capabilities, ensure that either:
 
 [This comment](https://github.com/palantir/palantir-java-format/issues/952#issuecomment-2575750610) explains why we 
 switched to a native image for the formatter. The startup time for the native image esp. in Intellij is >10x faster than
-spinning up a new process that does the formatting.
+spinning up a new JVM process that does the formatting.
 However, the throughput of running the native image for a large set of files (eg. running `./gradlew spotlessApply`) is 
 considerably slower (eg. 30s using the Java implementation vs 1m20s using the native image implementation). Therefore,
 when running the formatter from `spotlessApply` we will default to using the Java implementation (if the Java version >= 21).
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ Java 21 formatting capabilities, ensure that either:
 switched to a native image for the formatter. The startup time for the native image esp. in Intellij is >10x faster than
 spinning up a new process that does the formatting.
 However, the throughput of running the native image for a large set of files (eg. running `./gradlew spotlessApply`) is 
-considerably slower (eg. 30ms using the Java implementation vs 1m20s using the native image implementation). Therefore,
+considerably slower (eg. 30s using the Java implementation vs 1m20s using the native image implementation). Therefore,
 when running the formatter from `spotlessApply` we will default to using the Java implementation (if the Java version >= 21).
 
 
--- a/gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
+++ b/gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
@@ -18,6 +18,11 @@ package com.palantir.javaformat.gradle
 import nebula.test.IntegrationTestKitSpec
 import spock.lang.Unroll
 
+import java.nio.charset.StandardCharsets
+import java.nio.file.Path
+import java.util.stream.Collectors
+import java.util.stream.Stream
+
 class PalantirJavaFormatSpotlessPluginTest extends IntegrationTestKitSpec {
     /** ./gradlew writeImplClasspath generates this file. */
     private static final CLASSPATH_FILE = new File("build/impl.classpath").absolutePath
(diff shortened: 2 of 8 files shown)


gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
```diff
@@ -56,19 +103,58 @@ class PalantirJavaFormatSpotlessPluginTest extends IntegrationTestKitSpec {
...
-        "" | "Using the legacy palantir-java-formatter"
-        "palantir.native.formatter=true"  | "Using the native-image"
+        extraGradleProperties               | javaVersion   | expectedOutput
+        ""                                  | 21            | "Using the Java-based formatter"
+        "palantir.native.formatter=true"    | 21            | "Using the Java-based formatter"
+        "palantir.native.formatter=true"    | 17            | "Using the native-image formatter"
+
+    }
+
+    private static Iterable<File> getBuildPluginClasspathInjector() {
```
crogoz, line 122: should I move this to https://github.com/palantir/gradle-plugin-testing ? gradle-jdks needs to do the same thing in order to run on multiple Java versions.

README.md
```diff
@@ -182,6 +182,25 @@ shortcut.
...
+
+- the Gradle daemon and the Intellij Project SDK are set to Java 21
+- or that the gradle property `palantir.native.formatter=true`. This will run the formatter as a native image, 
+- independent of the Gradle daemon/Intellij project JDK version. 
+
+### Native image formatter
+
+[This comment](https://github.com/palantir/palantir-java-format/issues/952#issuecomment-2575750610) explains why we 
+switched to a native image for the formatter. The startup time for the native image esp. in Intellij is >10x faster than
+spinning up a new process that does the formatting.
```
felixdesouza: I think this is a tad confusing, in both cases you're spinning up a new process.

Correct me if I've misunderstood, but I believe you want something along the lines of:

> is >10x faster than spinning up a new JVM process that does the formatting

?

README.md
```diff
@@ -182,6 +182,25 @@ shortcut.
...
+However, the throughput of running the native image for a large set of files (eg. running `./gradlew spotlessApply`) is 
+considerably slower (eg. 30ms using the Java implementation vs 1m20s using the native image implementation). Therefore,
```
felixdesouza: 30 milliseconds vs 1 minute 20s?

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
felixdesouza: I don't understand what it's supposed to do?

felixdesouza: I suppose what I don't understand is why you need this at all? I thought nebula does this for you?

crogoz: I need to test how the task works against Java 17 and Java 21. So, I am using the `gradle-jdks automanagement workflow` to set-up the [JDKs](https://github.com/palantir/palantir-java-format/pull/1243/files#diff-fca5e8107f02ed08cfc2b20b592d3b4a73284c64a8749250d9506ba95cbdee16R79-R83) but then I also need to run [`./gradlew` instead of the classic `nebula runTask` ](https://github.com/palantir/palantir-java-format/pull/1243/files#diff-fca5e8107f02ed08cfc2b20b592d3b4a73284c64a8749250d9506ba95cbdee16R109). When using nebula `runTask` the plugin is injected automatically, but now [I need to set it on my own](https://github.com/palantir/palantir-java-format/pull/1243/files#diff-fca5e8107f02ed08cfc2b20b592d3b4a73284c64a8749250d9506ba95cbdee16R59). This is what `getBuildPluginClasspathInjector` does - it adds the files of the plugin to the test classpath.

I have the same setup in `gradle-jdks` for tests - so I was wondering if it makes sense to move this logic in  the testing plugin.

felixdesouza: oh, I thought the tooling api invoked gradlew, interesting!

you could move it to gradle-plugin-testing, it's a bit niche, but I don't see why not.

felixdesouza: I think a nicer way to do this, is probably to reuse some of the stuff in Nebula, it makes your life a lot easier I think:

* Use `ClasspathAddingInitScriptBuilder` to write the classpath in an init script form
* include the init script when running `gradlew`. i.e. `./gradlew <task> -i init-script`

I think the current approach - I also mentioned something similar in the testing plugin PRs - detracts from the test by adding a bunch of cruft that's not actually important to know for the test, but is unfortunately required for the test to pass.

felixdesouza: It also means, that you can hide all of this behind a `runGradlew` method and not have to pollute the test with it.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
```diff
@@ -26,20 +31,61 @@ class PalantirJavaFormatSpotlessPluginTest extends IntegrationTestKitSpec {
...
+            apply plugin: 'com.palantir.jdks.latest'
             
             dependencies {
                 palantirJavaFormat files(file("${CLASSPATH_FILE}").text.split(':'))
                 EXTRA_CONFIGURATION
             }
-        """.replace("EXTRA_CONFIGURATION", extraDependencies).stripIndent()
+            
+            javaVersions {
+                libraryTarget = JAVA_VERSION
```
felixdesouza: Can we just string interpolate here?

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
crogoz: addressed in https://github.com/palantir/palantir-java-format/pull/1243/commits/9cd9bc2b530f7d6473be7cb13ab01bea01d315f1

however, I still need to keep this method, if I use the `ClassLoader` to add all the files loaded by default - the test will fail [here](https://github.com/palantir/palantir-java-format/blob/de62f3b0b93c2d0e9764eb864ff693e1b52f440c/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L85)

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
```diff
@@ -16,30 +16,74 @@
...
     private static final NATIVE_IMAGE_FILE = new File("build/nativeImage.path")
     private static final NATIVE_CONFIG = String.format("palantirJavaFormatNative files(\"%s\")", NATIVE_IMAGE_FILE.text)
+    private static final String INIT_FILE_NAME = "init.gradle";
 
 
     @Unroll
-    def "formats with spotless when spotless is applied"(String extraGradleProperties, String expectedOutput) {
+    def "formats with spotless when spotless is applied"(String extraGradleProperties, String javaVersion, String expectedOutput) {
+        File initScript = new File(projectDir, INIT_FILE_NAME)
+        ClasspathAddingInitScriptBuilder.build(initScript, getBuildPluginClasspathInjector().toList())
```
felixdesouza: I think this should move into `runGradlewTasks` then you don't even need to have a constant. It's kinda similar to how the nebula thing does it, doesn't necessarily have to be a temp file, but I don't think it's anything specific to this particular test unlike `CLASSPATH_FILE` and `NATIVE_CONFIG` etc.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
felixdesouza: An aside: I think you could probably ship this method + `runTasksViaGradlew` into gradle-testing-plugin or a new thing. It'd just look very similar to `IntegrationSpec` or something.

felixdesouza (dismissed):

CRogers:
:+1:

PR #644, opened 2022-01-28
Handle empty implementation version

schlosna:
## Before this PR
From #643, encountered `NullPointerException` in `PalantirJavaFormatSettings.injectedVersionIsOutdated` below on IntelliJ IDEA 2022.1 EAP (`#IC-221.3427.89`) with palantir-java-format (2.16.0)

```
java.lang.NullPointerException
  at com.palantir.javaformat.intellij.PalantirJavaFormatSettings.injectedVersionIsOutdated(PalantirJavaFormatSettings.java:97)
  at com.palantir.javaformat.intellij.FormatterProvider.get(FormatterProvider.java:58)
  at com.palantir.javaformat.intellij.PalantirCodeStyleManager.format(PalantirCodeStyleManager.java:196)
  at com.palantir.javaformat.intellij.PalantirCodeStyleManager.formatInternal(PalantirCodeStyleManager.java:185)
  at com.palantir.javaformat.intellij.PalantirCodeStyleManager.reformatText(PalantirCodeStyleManager.java:110)
  at com.intellij.codeInsight.editorActions.smartEnter.SmartEnterProcessor.reformat(SmartEnterProcessor.java:38)
```

## After this PR
Fixes #643 

==COMMIT_MSG==
Handle empty implementation version
==COMMIT_MSG==

PR #1247, opened 2025-03-20
Fix pluginVersion discovery

crogoz:
## Before this PR

The logic of using the bundled/injected version doesn't work as expected. I think at some point the [logic that gets the version of the plugin ](https://github.com/palantir/palantir-java-format/compare/cr/fix-version?expand=1#diff-cfaa2a5a1981704685ebc0f2752eec605ccdb32fc40126b4cc05e1dfa27f900bL115) stopped returning a version - probably this was surfaced in [here](https://github.com/palantir/palantir-java-format/pull/644/files#diff-e7ad752a0f7a73390d1661fcc4aefc8038ca9564d30c6bd8520c5477276b557bR100) and now the implementation version is always `unknown`. This makes the logic to figure out whether to use the bundled impl or the injected impl always return the bundled impl - because the bundledImpl is an [Optional.empty](https://github.com/palantir/palantir-java-format/blob/d8845f8dba83a3f18c5c58a7e36d88bd3fe8fe97/idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatSettings.java#L106) hence we will always use the bundledImplementation.

Screenshot 2025-03-20 at 2 16 59 PM

## After this PR

Correctly figure out the version of the plugin.
Screenshot 2025-03-20 at 2 06 34 PM

==COMMIT_MSG==
Fix pluginVersion discovery
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1247.v2.yml
+++ b/changelog/@unreleased/pr-1247.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Fix pluginVersion discovery
+  links:
+  - https://github.com/palantir/palantir-java-format/pull/1247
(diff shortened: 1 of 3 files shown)


CRogers:
:+1:

PR #1306, opened 2025-05-23
Intellij uses the native image from `.gradle/palantir-java-formatter-caches`

crogoz:
## Before this PR

Unused artifact transforms are cleaned up by Gradle [every 7 days](https://docs.gradle.org/8.14/userguide/directory_layout.html#dir:gradle_user_home:configure_cache_cleanup:~:text=Created%20resources%3A,is%207%20days.). This could lead to situation - if people do not refresh the Gradle project in Intellij - where the formatter would fail to run because:

```
Caused by: java.io.IOException: Cannot run program "~/.gradle/caches/8.12.1/transforms/8b5bfec0f342b9bc0fee275f271594d7/transformed/palantir-java-format-native-2.63.0-nativeImage-macos_aarch64.bin.executable" (in directory "/~/.gradle/caches/8.12.1/transforms/8b5bfec0f342b9bc0fee275f271594d7/transformed"): error=2, No such file or directory
```
internal slack thread [here](https://pl.ntr/2t-)

## After this PR
- When we are writing the Intellij configuration files, we would firstly copy + move the nativeImage to our own cache (in `~/.gradle/palantir-java-formatter-caches`).
- the Intellij configuration files are updated using Gradle tasks that also get added as startParameters (similar logic to https://github.com/palantir/gradle-idea-configuration/blob/develop/gradle-idea-configuration/src/main/groovy/com/palantir/gradle/ideaconfiguration/IdeaConfigurationPlugin.java#L48)

==COMMIT_MSG==
Intellij uses the native image from `.gradle/palantir-java-formatter-caches`
==COMMIT_MSG==


crogoz:
adding `do not merge` until I finish testing the last commit

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/FileUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+package com.palantir.javaformat.gradle;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Comparator;
+import java.util.stream.Stream;
+
+public final class FileUtils {
```
crogoz, line 25: same as in gradle-jdks

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/NativeImageAtomicCopy.java
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.file.StandardOpenOption;
+import java.util.Optional;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+
+public class NativeImageAtomicCopy {
+
+    private static Logger logger = Logging.getLogger(PalantirJavaFormatIdeaPlugin.class);
+
+    public static URI copyToCacheDir(URI srcUri) {
```
crogoz: similar to the logic in gradle-jdks https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-setup/src/main/java/com/palantir/gradle/jdks/setup/GradleJdkInstallationSetup.java#L114 with a few modification eg. not copying a directory, but a file

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -53,13 +41,65 @@ public void apply(Project rootProject) {
 
         rootProject.getPlugins().apply(PalantirJavaFormatProviderPlugin.class);
         rootProject.getPluginManager().withPlugin("idea", ideaPlugin -> {
-            Configuration implConfiguration =
-                    rootProject.getConfigurations().getByName(PalantirJavaFormatProviderPlugin.CONFIGURATION_NAME);
-
-            Optional<Configuration> nativeImplConfiguration = maybeGetNativeImplConfiguration(rootProject);
-
-            configureLegacyIdea(rootProject, implConfiguration, nativeImplConfiguration);
```
crogoz, line 70: Removed the `configureLegacyIdea` logic completely. I see that we did that in `gradle-idea-configuration` as well but happy to bring it back if you think this might be problematic for some usecases.

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -76,87 +111,4 @@ private static Optional<Configuration> maybeGetNativeImplConfiguration(Project r
...
-
-    private static void configureIntelliJImport(
-            Project project, Configuration implConfiguration, Optional<Configuration> nativeImplConfiguration) {
-        // Note: we tried using 'org.jetbrains.gradle.plugin.idea-ext' and afterSync triggers, but these are currently
-        // very hard to manage as the tasks feel disconnected from the Sync operation, and you can't remove them once
-        // you've added them. For that reason, we accept that we have to resolve this configuration at
-        // configuration-time, but only do it when part of an IDEA import.
-        if (!Boolean.getBoolean("idea.active")) {
-            return;
-        }
```
crogoz, line 112: I am not sure here - do we still need to check the `idea.active` and run this when the `idea` plugin is applied ?

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/NativeImageAtomicCopy.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private static Path getGradleCacheNativeImagesDir() {
+        Path nativeImagesCacheDir = getGradleCacheDir().resolve("palantir-java-format-caches");
+        nativeImagesCacheDir.toFile().mkdirs();
+        return nativeImagesCacheDir;
+    }
+
+    private static Path getGradleCacheDir() {
+        return Path.of(Optional.ofNullable(System.getenv("GRADLE_USER_HOME"))
+                .orElseGet(() -> System.getenv("HOME") + "/.gradle"));
```
CRogers: I think we can use `project.getGradle().getGradleUserHomeDir()` instead

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
felixdesouza: let's drop it!

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -53,13 +41,60 @@ public void apply(Project rootProject) {
...
+                        task.getOutputFile().set(rootProject.file(".idea/palantir-java-format.xml"));
+                        task.getShouldCreateOutputIfAbsent().set(true);
+                    });
+
+            TaskProvider<UpdatePalantirJavaFormatIdeaXmlFile> updatePalantirJavaFormatIpr = rootProject
+                    .getTasks()
+                    .register("updatePalantirJavaFormatIpr", UpdatePalantirJavaFormatIdeaXmlFile.class, task -> {
+                        task.getOutputFile().set(rootProject.file(rootProject.getName() + ".ipr"));
+                        task.getShouldCreateOutputIfAbsent().set(false);
+                    });
```
felixdesouza: do we still need this? I don't have any `.ipr` files in my directories anymore

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -53,13 +41,60 @@ public void apply(Project rootProject) {
...
+                        task.getOutputFile().set(rootProject.file(".idea/workspace.xml"));
+                        task.getShouldCreateOutputIfAbsent().set(true);
+                    });
+
+            TaskProvider<UpdateWorkspaceXmlFile> updateWorkspaceIwsXml = rootProject
+                    .getTasks()
+                    .register("updateWorkspaceIwsXml", UpdateWorkspaceXmlFile.class, task -> {
+                        task.getOutputFile().set(rootProject.file(rootProject.getName() + ".iws"));
+                        task.getShouldCreateOutputIfAbsent().set(false);
+                    });
```
felixdesouza: similarly here, isn't this just the `idea` logic?

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
CRogers: I think we could probably drop the old way now. We've had the huge warning via baseline for ages now, and it's super deprecated.

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/UpdatePalantirJavaFormatIdeaXmlFile.java
```diff
@@ -0,0 +1,63 @@
...
+    @Optional
+    @InputFiles
+    public abstract ConfigurableFileCollection getNativeImageConfig();
+
+    @Optional
+    @OutputFile
+    public abstract RegularFileProperty getOutputFile();
+
+    @OutputDirectory
+    public abstract RegularFileProperty getCacheDir();
```
CRogers: This cache dir can have multiple different formatters in it right? If so, we should just have a single output that is the file we expect it to be in, otherwise Gradle will regard this task not up-to-date as extra stuff gets added to it.

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/UpdatePalantirJavaFormatIdeaXmlFile.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @TaskAction
+    public final void updateXml() {
+        List<URI> uris =
+                getImplementationConfig().getFiles().stream().map(File::toURI).collect(Collectors.toList());
+        java.util.Optional<URI> nativeUri = getNativeImageConfig().isEmpty()
+                ? java.util.Optional.empty()
+                : java.util.Optional.of(getNativeImageConfig().getSingleFile().toURI())
```
CRogers: I'd probably preferentially import `java.util.Optional` and fully qualify the `org.gradle.api.tasks.Optional` to make this a bit more readable.

CRogers: Instead of this, you can use `getImplementationConfig().getFiles().stream().findFirst().map(File::toUri)`

CRogers (approved):

PR #1312, opened 2025-06-04
Support for var in record patterns

### LINKED ISSUE #1309: Record patterns should be supported
```java
record X(int n) {}

void foo(Object x) {
  if (x instanceof X(var n)) {
    // ...
  }
}
```

compiles and works, but fails Spotless with

```
palantir-java-format(palantir-java-format) error: expected token: 'var'; generated n instead
```

Notes:
* I'm using Java 21, Gradle 8.14 and Spotless 7.0.3
* Switching from `var n` to `int n` causes the same failure
* Google-Java-Format is able to handle this syntax

emandirola:
## Before this PR
palantir fails when parsing source files with `var` inside a record pattern.
e.g.
```java
if (o instanceof Record(var i)) {}
switch (o) {
  case Record(var i):
}
```

## After this PR
Added support for `var` in record patterns.
The AST explicitly unsets the type when it sees `var`.
We may assume that's the reason `type` is `null`, so we just output "var".
Fixes #1309 

## Possible downsides?
The `type` is set to `null` when the token found is `var`, might change in the future?
There's a `declaredUsingVar` but it's false, might be broken (including java 24)

### DIFF
--- a/changelog/@unreleased/pr-1312.v2.yml
+++ b/changelog/@unreleased/pr-1312.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Support for var in record patterns
+  links:
+  - https://github.com/palantir/palantir-java-format/pull/1312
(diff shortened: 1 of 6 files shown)


CRogers:
:+1:

felixdesouza:
👍

PR #1313, opened 2025-06-04
Make palantir-java-format support spotless 6.22.0

kelvinou01:
## Before this PR

We are making [gradle-baseline](https://github.com/palantir/gradle-baseline) configuration cache friendly. The first step is to get `./gradlew classes --configuration-cache` to work to work on gradle-baseline. The command raised [a few errors in the build scan, which were resolved by bumping spotless from to 6.22.0](https://github.com/palantir/gradle-baseline/pull/3119). However, after bumping, `./gradlew classes --configuration-cache` raised these new errors in the build scan:

```
A problem occurred configuring root project '<redacted>-root'.
> A problem occurred configuring project ':<redacted>'.
   > Not allowed to resolve configuration ':palantirJavaFormat' at configuration time (https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time). Please upgrade your plugins and double-check your gradle scripts (see stacktrace)
```

The error actually comes from gradle-consistent-versions, [which enforces that configurations cannot be resolved at configuration time](https://github.com/palantir/gradle-consistent-versions/blob/9641411e4adc90c74159697a183f5dda1111dc0a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java#L233). We used to resolve the dependencies of the palantirJavaFormat [configuration](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.Configuration.html) at configuration time, and the bump to spotless has somehow surfaced these errors. 

## After this PR

By deferring dependency resolution to build time (the right thing to do), the error is resolved.

==COMMIT_MSG==
Make palantir-java-format support spotless 6.22.0, which is [transitively brought in from baseline to support the configuration cache](https://github.com/palantir/gradle-baseline/pull/3119)

==COMMIT_MSG==

## Possible downsides?

Note that palantir-java-format still doesn't support configuration cacheing by itself, due to some external calls done in NativeImageFormatProviderPlugin. Specifically, OperatingSystems::get calls `ldd`.


gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java
```diff
@@ -25,41 +25,37 @@
...
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.function.Supplier;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 
 public final class NativePalantirJavaFormatStep {
-    private static Logger logger = Logging.getLogger(NativePalantirJavaFormatStep.class);
+    private static final Logger logger = Logging.getLogger(NativePalantirJavaFormatStep.class);
```
FinlayRJW, line 34: The logger is used within the `State` class so should be created on that class - else logging will show it coming from `NativePalantirJavaFormatStep` which may be confusing

kelvinou01: As discussed in person, putting it in NativePalantirJavaFormatStep or State are both fine design decisions. Let's go with NativePalantirJavaFormatStep for now!

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -51,25 +51,36 @@ static final class State implements Serializable {
...
         }
 
         @SuppressWarnings("NullableProblems")
         FormatterFunc createFormat() {
-            return memoizedFormatter.get()::formatSourceReflowStringsAndFixImports;
+            return input -> {
+                try {
+                    // Only resolve the jars and compute the signature at execution time!
+                    Iterable<File> jars = jarsSupplier.get();
+                    this.jarsSignature = FileSignature.signAsSet(jars);
```
felixdesouza, line 78: I think it's worth figuring out how often this is called. It's the only thing that will be repeatedly called, and jars can be big, so you don't want to spend a ton of time doing jar hashing repeatedly and unnecessarily

changelog/@unreleased/pr-1313.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Support configuration cache
```
felixdesouza: You should be explicit about what tasks you're supporting via configuration cache. 

In any case, the changelog is not correct, as this PR is only for spotless 6.22.0 which is *part* of supporting configuration cache, but not all.

felixdesouza:
Not sure whether spotless is a dependency here, if so, it's worth bumping it. If not, worth testing it, there's probably a test that has spotless/gradle-baseline-spotless, that we can latch onto and bump the version of spotless to

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
kelvinou01: Running spotlessApply on the whole project with multiple modified files only runs `createFormat` once, so I think there's no perf concern here. 

Also, hashing is needed to construct the State so I don't think we can get rid of it here.

changelog/@unreleased/pr-1313.v2.yml
kelvinou01: Yep, I'll make the changelog here clearer!

kelvinou01:
> Not sure whether spotless is a dependency here, if so, it's worth bumping it. If not, worth testing it, there's probably a test that has spotless/gradle-baseline-spotless, that we can latch onto and bump the version of spotless to

@felixdesouza  As discussed in person, let's make spotless a `constraint`.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
felixdesouza: I wasn't suggesting getting rid of hashing, just not doing it repeatedly i.e. how many times is `createFormat` called. 

I think there's a separate issue where we create this state once per *project*, so the more projects you have, you create a `FormatterFunc` and therefore repeatedly hash the contents. This exists before this change and this change doesn't make it worse. In the context of build perf, it's probably worth making this go away. I would mark this as a FLUP for this PR.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPluginTest.groovy
```diff
@@ -57,6 +57,10 @@ class PalantirJavaFormatSpotlessPluginTest extends IntegrationTestKitSpec {
                      classpath 'com.palantir.baseline:gradle-baseline-java:6.21.0'
                      classpath 'com.palantir.gradle.jdks:gradle-jdks:0.62.0'
                      classpath 'com.palantir.gradle.jdkslatest:gradle-jdks-latest:0.17.0'
+
+                     constraints {
```
felixdesouza: just to double check, if you reverted your changes, did you see the test fail?

kelvinou01: Unexpectedly, it doesn't break if we revert our changes. I am trying to come up with a test that does. 

I found out that the error that necessitated this PR is from gradle-consistent-versions [enforcing that configurations cannot be resolved at configuration time](https://github.com/palantir/gradle-consistent-versions/blob/9641411e4adc90c74159697a183f5dda1111dc0a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java#L233). Specifically, in pjf, it's the [native formatter](https://github.com/palantir/palantir-java-format/blob/00b08d2f471d66382d6c4cd2d05f56b6bb546ad3/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java#L44) that breaks this constraint.

I wrote a test which applies consistent-versions and runs spotlessApply with the native formatter (at okelvin/cc-test), but seems like the restriction is applied too late, and hence is not given the chance to trigger:

(some test logs i added, that show the restriction is applied too late)
```
|      Selected primary task 'spotlessApply' from project :
|      Using the native-image formatter
|      NativePalantirJavaFormatStep: Resolving the palantirJavaFormatNative Configuration
|      VersionsPropsPlugin: restricting Configuration resolution at configuration time
```

kelvinou01: I've added the SupportsSpotless622 test, which fails for develop "Not allowed to resolve configuration ':palantirJavaFormat' at configuration time", but passes with this PR.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
```diff
@@ -0,0 +1,100 @@
...
+
+    GradlewExecutor(@NonNull File projectDir) {
+        this.projectDir = projectDir
+    }
+
+    private static List<File> getBuildPluginClasspathInjector() {
+        return PluginUnderTestMetadataReading.readImplementationClasspath();
+    }
+
+    private static Iterable<File> getPluginClasspathInjector(Path path) {
```
felixdesouza: this isn't used?

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
```diff
@@ -0,0 +1,100 @@
...
+        ProcessBuilder processBuilder = getProcessBuilder(tasks)
+        Process process = processBuilder.start()
+        String output = readAllInput(process.getInputStream())
+        process.waitFor(1, TimeUnit.MINUTES)
+        return new GradlewExecutionResult(process.exitValue(), output)
+    }
+
+    final class GradlewExecutionResult {
+
+        final Boolean success
```
felixdesouza: nit: boolean

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
```diff
@@ -0,0 +1,89 @@
...
+ * IntegrationTestKitSpec currently loads more than what it needs into the classpath
+ * https://github.com/nebula-plugins/nebula-test/blob/c5d3af9004898276bde5c68da492c6b0b4c5facc/src/main/groovy/nebula/test/IntegrationTestKitBase.groovy#L136
+ * This means if we run a test with IntegrationTestKitSpec's runner, the Formatter is loaded onto the classpath eagerly.
+ * If the test applies the PalantirJavaFormatPlugin, it complains that the Formatter is erroneously loadable
+ * https://github.com/palantir/palantir-java-format/blob/00b08d2f471d66382d6c4cd2d05f56b6bb546ad3/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L83.
+ * To be clear, this complaint is entirely a result of the IntegrationTestKitSpec loading too many things onto classpath.
+ *
+ * As a workaround, this runner uses the classpath produced by Gradle Test Kit in plugin-under-test-metadata.properties.
+ * This classpath only contains the dependencies required by the plugin, as well as the plugin itself.
+ * This means no more eager loading of Formatters onto the classpath, no more complaints from PalantirJavaFormatStep.
```
felixdesouza:
```suggestion
 * {@link IntegrationTestKitSpec} currently loads <a href="https://github.com/nebula-plugins/nebula-test/blob/c5d3af9004898276bde5c68da492c6b0b4c5facc/src/main/groovy/nebula/test/IntegrationTestKitBase.groovy#L136"> more than what it needs into the classpath</a>.
 * This means if we run a test with {@link IntegrationTestKitSpec}'s runner, the {@link Formatter} is on the build's classpath by virtue of being in the test's classpath.
 * If the test applies the {@link PalantirJavaFormatPlugin}, it complains that the {@link Formatter} is <a href="https://github.com/palantir/palantir-java-format/blob/00b08d2f471d66382d6c4cd2d05f56b6bb546ad3/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L83">erroneously loadable</a>.
 * To be clear, this complaint is entirely a result of the {@link IntegrationTestKitSpec} loading too many things onto classpath since it doesn't know what the exact plugin classpath is.
 * As a workaround, this runner uses the classpath produced by Gradle Test Kit in {@code plugin-under-test-metadata.properties}.
 * This classpath only contains the dependencies required by the plugin, as well as the plugin itself.
 * This means that even if we put the formatter on the {@code testClassPath}, it won't leak through to the Gradle build under test and subsequently no error from {@link PalantirJavaFormatStep}.
```

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+import nebula.test.IntegrationTestKitSpec
+
+/**
+ * When we were getting gradle-baseline to support the configuration cache, spotless had some poorly written tasks
+ * which caused issues with the configuration cache.
+ *
+ * Bumping spotless to 6.22.0 fixed this, but revealed a new error — the palantirJavaFormat configuration was
+ * being resolved eagerly:
+ * https://github.com/palantir/palantir-java-format/blob/b7b5995df3be690780939c0d0cb2ec49b99c68c8/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java#L45
```
felixdesouza:
```suggestion
 * Bumping spotless to 6.22.0 fixed this, but revealed a new error — the palantirJavaFormat configuration was
 * being <a href="https://github.com/palantir/palantir-java-format/blob/b7b5995df3be690780939c0d0cb2ec49b99c68c8/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java#L45">resolved eagerly</a>
```

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,90 @@
...
+ */
+package com.palantir.javaformat.gradle
+
+import nebula.test.IntegrationTestKitSpec
+
+/**
+ * When we were getting gradle-baseline to support the configuration cache, spotless had some poorly written tasks
+ * which caused issues with the configuration cache.
+ *
+ * Bumping spotless to 6.22.0 fixed this, but revealed a new error — the palantirJavaFormat configuration was
```
felixdesouza:
```suggestion
 * Bumping spotless to 6.22.0 fixed this, but revealed a new error — the {@code palantirJavaFormat} configuration was
```

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,90 @@
...
+            project.getTasks().getByName("spotlessJava")
+        """.stripIndent()
+
+        when:
+        def result = executor.runGradlewTasks('classes', '--info')
+
+        then:
+        assert result.success
+        println(result.standardOutput)
+        result.standardOutput.contains('BUILD SUCCESSFUL')
```
felixdesouza: you shouldn't need to do this

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            // This forces the realization of the spotlessJava task, creating the spotless steps. 
+            // If any configurations are eagerly resolved in the spotless steps, 
+            // consistent-versions should catch it and throw here. 
+            project.getTasks().getByName("spotlessJava")
+        """.stripIndent()
+
+        when:
+        def result = executor.runGradlewTasks('classes', '--info')
```
felixdesouza, line 82: is it worth running a format? idk, or is just forcing the realisation of the task enough

kelvinou01: Yep! It was the realization of the task that caused the original error in gradle-baseline. Also, this test breaks when we remove our fixes.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
kelvinou01: yep! good catch.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
kelvinou01: gotcha

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
kelvinou01: Got it, this is how you do proper formatting for links in javadocs.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
kelvinou01: Yep, the `assert result.success` should be sufficient.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java
```diff
@@ -25,41 +25,37 @@
...
+        State(Supplier<File> supplier) {
+            this.execSupplier = supplier;
         }
 
         String format(ProcessRunner runner, String input) throws IOException, InterruptedException {
-            List<String> argumentsWithPathToExe =
-                    List.of(pathToExe.getOnlyFile().getAbsolutePath(), "--palantir", "-");
+            File execFile = execSupplier.get();
+            logger.info("Using native-image at {}", execFile);
+            File signedFile = FileSignature.signAsSet(execFile).getOnlyFile();
```
CRogers: We still need to have this signature as part of the serialised state otherwise the task will be incorrectly up-to-date. Which means it needs to be a field still that's set at the "correct" time. I think we could make our own `writeObject`  method here to do this at the correct time.

CRogers: Eh, just saw `PalantirJavaFormatStep` - maybe we just need to do the same thing and set the field in `createFormat`. This whole setup definitely needs some comments, maybe that what #1327 would do, except we probably need a note to look in `PalantirJavaFormatStep` from this native one for some more explaination.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,87 @@
...
+    def setup() {
+        definePluginOutsideOfPluginBlock = true
+        keepFiles = true
+        executor = new GradlewExecutor(projectDir)
+    }
+
+    def "PalantirJavaFormatPlugin works with spotless 6.22.0"() {
+        // language=Gradle
+        buildFile << """
+        buildscript {
```
CRogers: I would indent this block by one as we're using `.stripIndent(true)` - will make it a bit more readable.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
```diff
@@ -0,0 +1,87 @@
...
+        buildFile << """
+            dependencies {
+                palantirJavaFormat files(file("${CLASSPATH_FILE}").text.split(':'))
+            }
+
+            // This forces the realization of the spotlessJava task, creating the spotless steps. 
+            // If any configurations are eagerly resolved in the spotless steps, 
+            // consistent-versions should catch it and throw here. 
+            project.getTasks().getByName("spotlessJava")
+        """.stripIndent()
```
CRogers: `.stripIndent(true)`

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
```diff
@@ -0,0 +1,86 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.javaformat.gradle
+
+import nebula.test.functional.internal.classpath.ClasspathAddingInitScriptBuilder
+import org.gradle.internal.impldep.org.eclipse.jgit.annotations.NonNull
```
CRogers: Don't use these `org.gradle.internal.impldep` types. They're Gradle's internal 3rd part deps. In this case you can just remove the `@NonNull` annotation entirely

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
```diff
@@ -0,0 +1,86 @@
...
+ * This means if we run a test with {@link IntegrationTestKitSpec}'s runner, the {@link Formatter} is on the build's classpath by virtue of being in the test's classpath.
+ * If the test applies the {@link PalantirJavaFormatPlugin}, it complains that the {@link Formatter} is <a href="https://github.com/palantir/palantir-java-format/blob/00b08d2f471d66382d6c4cd2d05f56b6bb546ad3/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L83">erroneously loadable</a>.
+ * To be clear, this complaint is entirely a result of the {@link IntegrationTestKitSpec} loading too many things onto classpath since it doesn't know what the exact plugin classpath is.
+ * As a workaround, this runner uses the classpath produced by Gradle Test Kit in {@code plugin-under-test-metadata.properties}.
+ * This classpath only contains the dependencies required by the plugin, as well as the plugin itself.
+ * This means that even if we put the formatter on the {@code testClassPath}, it won't leak through to the Gradle build under test and subsequently no error from {@link PalantirJavaFormatStep}.
+ */
+class GradlewExecutor {
```
CRogers: In fact, let's write this whole class in Java. There is no reason for it to be in groovy.

CRogers: Java is generally a better and more maintainable language

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
felixdesouza: You can add a thing to your intellij to exclude and not auto-import package prefixes! This makes it super easy to not get wrecked by checkstyle or these comments. 

Settings (⌘ + ,) > Editor > General > Auto Import > Exclude from auto import and completion

Screenshot 2025-06-23 at 14 34 26

You don't have to go back to the menu to keep doing that though however. Whenever you try to import a type and you get a menu, look for the alternatives proposed, you can then select the context menu by pressing the right arrow key, and then click the exclude button: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#exclude-from-auto-import

CRogers: We try to set this up in baseline for java. Maybe we should do the same for groovy?

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
kelvinou01: I am guessing the idea behind using Groovy for tests is that dynamic typing and syntactic sugar make scripting easier to write and read. For code that has to be maintained, we use Java.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/SupportsSpotless622.groovy
kelvinou01: Gotcha

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
CRogers: Yes, but also I think groovy for tests is just because we use netflix's nebula-lint, which uses spock, which requires groovy. In the old days we didn't have text blocks in Java so groovy was kinda essential as we needed an easy way to write blocks of gradle code in the tests - no Java has this feature we could probably avoid groovy.

gradle-palantir-java-format/src/test/groovy/com/palantir/javaformat/gradle/GradlewExecutor.groovy
kelvinou01: Thank you for the tip @felixdesouza @CRogers ! This sounds like a simple error prone as well, but unfortunately this doesn't catch mistakes in groovy.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/NativePalantirJavaFormatStep.java
kelvinou01: Yep, good catch! Let's keep the signature as a field in NativePalantirJavaFormatStep, so up to date checking can occur. 

Let's add a comment to this class in https://github.com/palantir/palantir-java-format/pull/1327

kelvinou01: I am wondering whether the up-to-date checking can be tested, but it seems complicated to set up for a small upside

kelvinou01:
Just a heads up, this is ready for re-review! @CRogers @felixdesouza

CRogers (approved):

PR #1327, opened 2025-06-18
Explain why we need to hash the palantir-java-format jar as part of SpotlessTask's state. 

kelvinou01:
## Before this PR

FLUP to https://github.com/palantir/palantir-java-format/pull/1313#discussion_r2137623833

~~In `PalantirJavaFormatStep`, we currently sign the palantir-java-format jar file, and serialize that signature as part of the `State`. I think [the `State` is *only* used to create the format function](https://github.com/diffplug/spotless/blob/52654ef8c4a6191d983b10a2370d53b1ca023f7d/lib/src/main/java/com/diffplug/spotless/FormatterStepImpl.java#L80), and nothing else like up-to-date checking. However, seems like the signature doesn't do anything, and just lives as a serializable field in `State`.~~

~~If I am correct, this is wasted computation, repeated once per subproject in your gradle build.~~

Seems like I was wrong. This hashing is both 1. efficient and 2. necessary. Let's add a comment explaning our findings. 

## After this PR

Explain why we need to hash the palantir-java-format jar as part of SpotlessTask's state. #1327

==COMMIT_MSG==
Explain why we need to hash the palantir-java-format jar as part of SpotlessTask's state
==COMMIT_MSG==


gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -53,34 +49,19 @@ static final class State implements Serializable {
         @SuppressWarnings({"unused", "FieldCanBeLocal"})
         private FileSignature jarsSignature;
```
CRogers, line 65: This doesn't seem to be used anywhere - can we just remove it?

CRogers: Even before it was not used - I don't get it.

CRogers: Oh, it was created in createFormat() - seems very odd.

CRogers: Ok, so I looked a bit deeper, and spotless has created this crazy system using Java Serialisation (which Gradle sort of understands) and this eventually becomes part of the input to `@Input SpotlessTask#getSteps`, so this is a Gradle input. We probably want to keep something here that represents the hashes of all files, so if this changed the task would not be up-to-date.

This is called signature, but it's not actually making a cryptographic signature - it's just [doing the name/size/hash](https://github.com/diffplug/spotless/blob/228eb10af382b19e130d8d9479f7a95238cb4358/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L181-L186). The hash is [sha-256](https://github.com/diffplug/spotless/blob/228eb10af382b19e130d8d9479f7a95238cb4358/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L151).

Spotless also has [caching](https://github.com/CRogers/spotless/blob/228eb10af382b19e130d8d9479f7a95238cb4358/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L138-L143) so this won't be recalculated multiple times.

All in all I think we should leave this. Maybe make so notes about what we've learned here as a comment at most

kelvinou01: Yep, I've added a comment explaining what we've found. 

> Gradle understands Java's `Serializable`, and uses it to invalidate `@Input`s to tasks.

Is this claim correct? I'm not sure where's the source code/docs. Is it [this](https://docs.gradle.org/current/userguide/configuration_cache.html#:~:text=Gradle%20uses%20an%20optimized%20serialization,simple%20state%20or%20supported%20types.)?

CRogers (dismissed):
Thanks for adding this explanation

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -30,7 +30,8 @@ public final class PalantirJavaFormatStep {
 
     private static final String IMPL_CLASS = "com.palantir.javaformat.java.Formatter";
 
-    private PalantirJavaFormatStep() {}
+    private PalantirJavaFormatStep() {
+    }
```
CRogers: Just need to fix the formatting issues like this and the too long comments.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -49,7 +50,16 @@ static final class State implements Serializable {
         @SuppressWarnings("unused")
         private final String stepName = NAME;
 
-        // Kept for state serialization purposes.
+        // Spotless' `FormatterStepImpl` implements Java's `Serializable` interface in a weird way:
```
kelvinou01, line 52: Add a similar explanation for NativePalantirJavaFormatStep

CRogers (dismissed):

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java
```diff
@@ -75,7 +86,11 @@ FormatterFunc createFormat() {
                 try {
                     // Only resolve the jars and compute the signature at execution time!
                     Iterable<File> jars = jarsSupplier.get();
+
+                    // Not a performance issue, as Spotless caches this
+                    // https://github.com/CRogers/spotless/blob/228eb10af382b19e130d8d9479f7a95238cb4358/lib/src/main/java/com/diffplug/spotless/FileSignature.java#L138-L143
```
felixdesouza: nit: can this point to `diffplug`?

kelvinou01: gotcha!

CRogers (approved):

PR #1338, opened 2025-06-30
Update to use the new configuration cache safe `GradleOperatingSystem`

FinlayRJW:
## Before this PR
https://github.com/palantir/gradle-baseline/pull/3140 was being blocked by PJF using `OperatingSystem` to run commands during the configuration phase via process builder. This break configuration cache and prevents merging.

## After this PR
https://github.com/palantir/gradle-utils/pull/383 resolved this by introducing `GradleOperatingSystem` a configuration cache safe way of getting the operating system, this PR updates PJF to use the new `GradleOperatingSystem` class fixing the issues allowing for baseline to merge

==COMMIT_MSG==
Update to use the new configuration cache safe `GradleOperatingSystem`
==COMMIT_MSG==


gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -69,8 +75,9 @@ public void apply(Project rootProject) {
                 .register("palantir-java-format", dep -> dep.atLeastVersion(MIN_IDEA_PLUGIN_VERSION));
     }
 
-    private static Optional<Configuration> maybeGetNativeImplConfiguration(Project rootProject) {
-        return NativeImageFormatProviderPlugin.isNativeImageConfigured(rootProject)
+    private static Optional<Configuration> maybeGetNativeImplConfiguration(
+            Project rootProject, GradleOperatingSystem operatingSystem) {
```
kelvinou01: Ideally a method should take in the smallest API it needs. Let's take in `Provider<OperatingSystem>` here, rather than `GradleOperatingSystem`?

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -38,10 +39,14 @@
...
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.tasks.Nested;
 import org.gradle.plugins.ide.idea.model.IdeaModel;
 import org.xml.sax.SAXException;
 
-public final class PalantirJavaFormatIdeaPlugin implements Plugin<Project> {
+public abstract class PalantirJavaFormatIdeaPlugin implements Plugin<Project> {
+
+    @Nested
```
kelvinou01, line 49: Nice! Didn't know about this beforehand.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -54,26 +60,26 @@ public void apply(Project rootProject) {
...
-                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem));
+                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem.get()));
             transformSpec.getTo().attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "executable-nativeImage");
         });
     }
 
-    public static boolean isNativeImageConfigured(Project project) {
-        return isNativeImageSupported() && isNativeFlagEnabled(project);
+    public static boolean isNativeImageConfigured(Project project, GradleOperatingSystem operatingSystem) {
+        return isNativeImageSupported(operatingSystem) && isNativeFlagEnabled(project);
```
kelvinou01: nit: let's put `isNativeFlagEnabled(project)` in front here? seems like it's the cheaper check of the two (no external calls), so if that's false, we can short circuit

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -54,26 +60,26 @@ public void apply(Project rootProject) {
...
                     .getFrom()
-                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem));
+                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem.get()));
             transformSpec.getTo().attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "executable-nativeImage");
         });
     }
 
-    public static boolean isNativeImageConfigured(Project project) {
-        return isNativeImageSupported() && isNativeFlagEnabled(project);
+    public static boolean isNativeImageConfigured(Project project, GradleOperatingSystem operatingSystem) {
```
kelvinou01: Same here, let's take take in `Provider<OperatingSystem>`?

kelvinou01 (changes requested):

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
FinlayRJW: updated to take `Provider<OperatingSystem>` all over

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
FinlayRJW: yep good idea switched

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
FinlayRJW: yeah it's super useful!

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
FinlayRJW: good idea updated thanks!

kelvinou01 (dismissed):
Nice! Removing `merge when ready` so someone else can take a look at this before we merge

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -54,26 +61,26 @@ public void apply(Project rootProject) {
...
-    public static boolean isNativeImageConfigured(Project project) {
-        return isNativeImageSupported() && isNativeFlagEnabled(project);
+    public static boolean isNativeImageConfigured(Project project, Provider<OperatingSystem> operatingSystem) {
+        return isNativeFlagEnabled(project) && isNativeImageSupported(operatingSystem);
     }
 
-    private static boolean isNativeImageSupported() {
-        OperatingSystem os = OperatingSystem.get();
+    private static boolean isNativeImageSupported(Provider<OperatingSystem> operatingSystem) {
+        OperatingSystem os = operatingSystem.get();
```
felixdesouza: ~we should just call `operatingSystem` inside here, it's a field.~

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -69,8 +77,9 @@ public void apply(Project rootProject) {
                 .register("palantir-java-format", dep -> dep.atLeastVersion(MIN_IDEA_PLUGIN_VERSION));
     }
 
-    private static Optional<Configuration> maybeGetNativeImplConfiguration(Project rootProject) {
-        return NativeImageFormatProviderPlugin.isNativeImageConfigured(rootProject)
+    private static Optional<Configuration> maybeGetNativeImplConfiguration(
+            Project rootProject, Provider<OperatingSystem> operatingSystem) {
+        return NativeImageFormatProviderPlugin.isNativeImageConfigured(rootProject, operatingSystem)
```
felixdesouza: ~Same thing here, no real need to pass it in as a parameter, as long as this method isn't static anymore~

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
```diff
@@ -33,14 +35,14 @@ final class SpotlessInterop {
...
         SpotlessExtension spotlessExtension = project.getExtensions().getByType(SpotlessExtension.class);
-        spotlessExtension.java(java -> java.addStep(addSpotlessJavaFormatStep(project)));
+        spotlessExtension.java(java -> java.addStep(addSpotlessJavaFormatStep(project, operatingSystem)));
     }
 
-    static FormatterStep addSpotlessJavaFormatStep(Project project) {
+    static FormatterStep addSpotlessJavaFormatStep(Project project, Provider<OperatingSystem> operatingSystem) {
 
-        if (NativeImageFormatProviderPlugin.isNativeImageConfigured(project)
+        if (NativeImageFormatProviderPlugin.isNativeImageConfigured(project, operatingSystem)
```
felixdesouza: I see now

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -54,26 +61,26 @@ public void apply(Project rootProject) {
...
-                                getExtension(operatingSystem))));
+                                getExtension(operatingSystem.get()))));
             });
             conf.getAttributes().attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, "executable-nativeImage");
         });
         rootProject.getDependencies().registerTransform(ExecutableTransform.class, transformSpec -> {
             transformSpec
                     .getFrom()
-                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem));
+                    .attribute(ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE, getExtension(operatingSystem.get()));
```
felixdesouza: you're realising the operating system thing here, meaning it will always get called even if you're not going to resolve this configuration. You should use `getFrom().attributeProvider(...)`

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
felixdesouza: hmm, curious whether this get is avoidable where we are?

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -54,26 +61,26 @@ public void apply(Project rootProject) {
                         .create(String.format(
                                 "com.palantir.javaformat:palantir-java-format-native:%s:nativeImage-%s_%s@%s",
                                 implementationVersion,
-                                operatingSystem.uiName(),
+                                operatingSystem.get().uiName(),
                                 Architecture.get().uiName(),
-                                getExtension(operatingSystem))));
+                                getExtension(operatingSystem.get()))));
```
felixdesouza: you want to use `addAllLater` here:

```java
conf.defaultDependencies(deps -> {
  deps.addAllLater(operatingSystem.map(os -> {
    rootProject.getDependencies().create(String.format(...));
  }))
});
```

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
FinlayRJW: I guess we could return a `Provider<Boolean>` but would we not have to realise that immediately above when we do: 

`!isNativeImageConfigured(rootProject, operatingSystem))`

FinlayRJW: I guess you could do:

```
isNativeImageConfigured(rootProject, operatingSystem).map(isConfigured -> {
            if (!isConfigured) {
                log.info("Skipping native image configuration as it is not supported on this platform");
                return null;
            }
....
```

but is that any better?

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
```diff
@@ -33,14 +35,14 @@ final class SpotlessInterop {
 
     private SpotlessInterop() {}
 
-    static void addSpotlessJavaStep(Project project) {
+    static void addSpotlessJavaStep(Project project, Provider<OperatingSystem> operatingSystem) {
         SpotlessExtension spotlessExtension = project.getExtensions().getByType(SpotlessExtension.class);
-        spotlessExtension.java(java -> java.addStep(addSpotlessJavaFormatStep(project)));
+        spotlessExtension.java(java -> java.addStep(addSpotlessJavaFormatStep(project, operatingSystem)));
```
felixdesouza: worth commenting that this is okay because it happens in an `afterEvaluate`.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
```diff
@@ -37,12 +43,13 @@ public void apply(Project rootProject) {
                 rootProject == rootProject.getRootProject(),
                 "May only apply com.palantir.java-format-provider to the root project");
 
-        if (!isNativeImageConfigured(rootProject)) {
+        Provider<OperatingSystem> operatingSystem = getOs().getOperatingSystem();
+        if (!isNativeImageConfigured(rootProject, operatingSystem)) {
             log.info("Skipping native image configuration as it is not supported on this platform");
             return;
```
felixdesouza: don't think there is any harm in deleting this, it doesn't actually do anything by itself

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPlugin.java
```diff
@@ -16,10 +16,15 @@
...
 
 import com.google.common.collect.ImmutableList;
+import com.palantir.platform.GradleOperatingSystem;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.tasks.Nested;
+
+public abstract class PalantirJavaFormatSpotlessPlugin implements Plugin<Project> {
+    @Nested
+    protected abstract GradleOperatingSystem getOs();
```
felixdesouza: I kinda hate having to pull this here, just to pass it in.

I have a slightly crazier idea that I think works a bit better:

Have a `SpotlessInterop` implement `Action<JavaExtension>`.

```java
public abstract class SpotlessInterop implements Action<JavaExtension> {

  @Nested
  protected abstract GradleOperatingSystem getOs();

  @Inject
  protected abstract ConfigurationContainer getConfigurations();

  public SpotlessInterop() {}

  public void execute(JavaExtension java) {
    java.addStep(spotlessJavaFormatStep());
  }

  private FormatterStep spotlessJavaFormatStep() {
    // same as before but just use the methods instead of fields
  } 
}
```

Usage then becomes:

```java
SpotlessExtension spotlessExtension = project.getExtensions().getByType(SpotlessExtension.class);
// get from root project to have the right configurations
SpotlessInterop spotlessInterop = project.getRootProject().getObjects().newInstance(SpotlessInterop.class);
project.getPluginManager().withPlugin("java", _plugin -> {
  SPOTLESS_PLUGINS.forEach(
    spotlessPluginId -> project.getPluginManager().withPlugin(spotlessPluginId, _spotlessPlugin -> {
      spotlessExtension.java(spotlessInterop);
    }));
});
```

Benefits being it's relatively self contained,

FinlayRJW: I've converted to this format but needed to keep the project around to get the `JavaFormatExtension` and run `isNativeImageConfigured` - on the whole I do think it is neater though:

```java
public abstract class SpotlessInterop implements Action<JavaExtension> {
    private static final Logger logger = Logging.getLogger(SpotlessInterop.class);

    private final Project project;

    @Nested
    protected abstract GradleOperatingSystem getOs();

    @Inject
    protected abstract ConfigurationContainer getConfigurations();

    @Inject
    public SpotlessInterop(Project project) {
        this.project = project;
    }

    @Override
    public void execute(JavaExtension java) {
        java.addStep(spotlessJavaFormatStep());
    }

    private FormatterStep spotlessJavaFormatStep() {...}
```

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
FinlayRJW: removed

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/NativeImageFormatProviderPlugin.java
FinlayRJW: updated to:

```java
attributeProvider(
                            ArtifactTypeDefinition.ARTIFACT_TYPE_ATTRIBUTE,
                            operatingSystem.map(NativeImageFormatProviderPlugin::getExtension));
```

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/PalantirJavaFormatSpotlessPlugin.java
```diff
@@ -28,10 +29,14 @@ public class PalantirJavaFormatSpotlessPlugin implements Plugin<Project> {
     public void apply(Project project) {
         project.getRootProject().getPluginManager().apply(PalantirJavaFormatProviderPlugin.class);
 
+        SpotlessInterop spotlessInterop =
+                project.getRootProject().getObjects().newInstance(SpotlessInterop.class, project.getRootProject());
```
kelvinou01: nit: let's extract `project.getRootProject()` to a variable? 

Also, maybe `project.getObjects()` does the same thing.

kelvinou01 (commented):
Nice! Excited to see this go in.

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
CRogers: As discussed in person, we can make a `@Nested NativeImageSupport` type that contains these methods along with a `@Nested GradleOperatingSystem` and `@Inject ProviderFactory` to avoid having to pass this provider around everywhere

FinlayRJW: so I've done this and made a `NativeImageSupport` type - currently it has a method `boolean isNativeImageConfigured` would it be better to make the is a `Provider<Boolean>` or does it not really matter?

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatIdeaPlugin.java
```diff
@@ -34,14 +34,23 @@
...
+import org.gradle.api.artifacts.ConfigurationContainer;
+import org.gradle.api.tasks.Nested;
 import org.gradle.plugins.ide.idea.model.IdeaModel;
 import org.xml.sax.SAXException;
 
-public final class PalantirJavaFormatIdeaPlugin implements Plugin<Project> {
+public abstract class PalantirJavaFormatIdeaPlugin implements Plugin<Project> {
+
+    @Nested
+    protected abstract NativeImageSupport getNativeImageConfigured();
```
CRogers: `getNativeImageSupport`

gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/PalantirJavaFormatPlugin.java
```diff
@@ -46,7 +50,7 @@ public void apply(Project project) {
             // TODO(dfox): in the future we may want to offer a simple 'format' task so people don't need to use
             // spotless to try out our formatter
             project.getTasks().register("formatDiff", FormatDiffTask.class, task -> {
-                if (NativeImageFormatProviderPlugin.isNativeImageConfigured(project)) {
+                if (getNativeImageConfigured().isNativeImageConfigured()) {
                     task.getNativeImage().fileProvider(getNativeImplConfiguration(project));
```
felixdesouza: somehow my comment disappeared, or I never submitted, it, but similar thing here, `task.getNativeImage().fileProvider()` can take a provider

felixdesouza: so very similar to the `attributeProvider` thing you did

FinlayRJW: I'm not sure if I've miss-understood this but `getNativeImageConfigured().isNativeImageConfigured()` returns a `Boolean` and not a `Provider` so I guess you could do something like:

```java
                task.getNativeImage()
                        .fileProvider(getNativeImplConfiguration(project).map(file -> {
                            if (getNativeImageSupport().isNativeImageConfigured()) {
                                return file;
                            }
                            return null;
                        }));
```

but is that any better?

FinlayRJW: actually that doesnt seem to work I get 
```
Could not determine the dependencies of task ':formatDiff'.
> Failed to query the value of task ':formatDiff' property 'nativeImage'.
   > Could not resolve all files for configuration ':palantirJavaFormatNative'.
      > Cannot resolve external dependency com.palantir.javaformat:palantir-java-format-native:null because no repositories are defined.
        Required by:
            root project :
```
in the test

felixdesouza:
👍

CRogers:
:+1:

felixdesouza:
I think there are probably nicer ways to get this in i.e. using attributes and module metadata, so we can really just resolve / run programs at the last possible second. I might throw something up at a later point, but in its current form it's pretty hard to avoid

PR #1384, opened 2025-07-29
bump spotless to version to `7.2.1`

FinlayRJW:
## Before this PR
See baseline PR https://github.com/palantir/gradle-baseline/pull/3197

We where seeing a bug where using spotless `6.22.0` and `scalafmt` would error:

```
A problem occurred configuring root project 'foundry-spark-root'.
> A problem occurred configuring project ':foundry-spark'.
   > You need to add a repository containing the '[org.scalameta:scalafmt-core_2.12:3.6.1]' artifact in 'foundry-spark/build.gradle'.
     E.g.: 'repositories { mavenCentral() }'
```

I believe this was due to spotless not being lazy enough / fully supporting configuration cache - this was fixed in version [`7.0.0`](https://github.com/diffplug/spotless/releases/tag/gradle%2F7.0.0)

## After this PR
Use the latest version of spotless to fix the bug. 

==COMMIT_MSG==
bump spotless to version to `7.2.1`
==COMMIT_MSG==

## Possible downsides?
The update order is important, if we release this before baseline then it wont merge and visa versa I think the best order is:
- Release this version of PJF 
- update baseline to bump the version of spotless and set this version of PJF as the minimum 

This will mean that the update to this version of PJF will be blocked unless you have the new version of baseline but the update of baseline wont be blocked as it will contain the correct minimum version of PJF


CRogers:
:+1:

PR #1419, opened 2025-09-25
Format text blocks + Intellij reflows strings

### LINKED ISSUE #930: Text block get formatted ugly
## What happened?

this tool formats text block in not ideal format.

see original verion.
```java
    @Update("""
    update master_data_manager
    set version = version + 1, update_time = now(), update_by = #{userId}
    where site_id = #{siteId} and table_name = #{tableName}
    """)
```

output 

```java
    @Update(
            """
    update master_data_manager
    set version = version + 1, update_time = now(), update_by = #{userId}
    where site_id = #{siteId} and table_name = #{tableName}
    """)
```

keep the same as original.

### LINKED ISSUE #1331: Destructive formatting of multiline string
The mulit-line string formatting is broken.
This multi-line string
```
      """
          <div class="member-signature"><span class="annotations">@Generated(value="SomeGeneratedName",
                     date="a date",
                     comments="some comment about the method below")
          </span><span class="modifiers">public static</span>&nbsp;<span class="return-typ\
          e">void</span>&nbsp;<span class="element-name">annotatedMethod</span><wbr><span \
          class="parameters">(int&nbsp;i,
           java.lang.String&nbsp;s,
           boolean&nbsp;b)</span></div>"""
```
gets formatted to
```
        """
<div class="member-signature"><span class="annotations">@Generated(value="SomeGeneratedName",
date="a date",
comments="some comment about the method below")
</span><span class="modifiers">public static</span>&nbsp;<span class="return-typ\
e">void</span>&nbsp;<span class="element-name">annotatedMethod</span><wbr><span \
class="parameters">(int&nbsp;i,
java.lang.String&nbsp;s,
boolean&nbsp;b)</span></div>""",
```
As eagle-eyed persons will notice, this destroys the indentation completely.
From the looks of it, this formatting is done using a simple trim() on each line.
This should be fixed since this is a destructive error where the content is actually ruined.

crogoz:
## Before this PR

- Text blocks were not formatted. 
- Intellij would not wrap long strings/would not format textBlocks. Different behaviour when running the formatter via the Intellij plugin vs. the gradle task.

## After this PR

Fixes: https://github.com/palantir/palantir-java-format/issues/930
Fixes: https://github.com/palantir/palantir-java-format/issues/1331
- Text blocks get formatted, by adding the initial `"""` on the previous line, the text block gets reindented. 
- Same behaviour between the Intellij plugin & gradle task formatters.

Example of a re-format: https://github.com/palantir/gradle-guide/compare/cr/test-bump-pjf?expand=1
==COMMIT_MSG==
Format text blocks + Intellij reflows strings
==COMMIT_MSG==


palantir-java-format/src/main/java/com/palantir/javaformat/doc/Token.java
```diff
@@ -106,6 +107,11 @@ public void add(DocBuilder builder) {
 
     @Override
     protected float computeWidth() {
+        if (token.getTok().getOriginalText().startsWith(StringWrapper.TEXT_BLOCK_DELIMITER)) {
```
crogoz, line 110: before - when the textBlock+methods would exceed 120 characters (maxWidthLimit), the block would look like this:

```
    """
    hello %s this exceeds 120 characters nasddnsadalsdaslda
    """
        .formatted("world");
```

now: 

```
    """
    hello %s this exceeds 120 characters nasddnsadalsdaslda
    """.formatted("world");
```

palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java
```diff
@@ -49,14 +50,29 @@ public String call() throws FormatterException {
...
+            RangeSet<Integer> characterRanges = characterRanges(input);
+            Set<Range<Integer>> rangesToChange = characterRanges.asRanges();
+            List<Replacement> replacements = formatter.getFormatReplacements(input, rangesToChange);
+            if (parameters.reflowLongStrings()) {
+                String formattedText = Utils.applyReplacements(input, replacements);
+                // avoid trying to wrap the input unless the lines needing wrapping are part of the "characterRanges"
+                if (!StringWrapper.linesNeedWrapping(options.maxLineLength(), formattedText, characterRanges)) {
+                    return writeFormatReplacements(replacements);
+                }
+                String wrappedFormattedText = StringWrapper.wrap(options.maxLineLength(), formattedText, formatter);
```
crogoz, line 62: same behaviour now as with the gradle formatter task: https://github.com/palantir/palantir-java-format/blob/1cc116070dcbe74c14b5c870e66d395a6d0d9a71/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L72. 

The alternative would have been to directly call the same method in the [Intellij plugin](https://github.com/palantir/palantir-java-format/blob/develop/idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java#L105), however I think this would be a bit more optimized - because when there is no need to call the wrapper, we can simply return just the replacement ranges. Worst case happens here where we have to send back the full document range to be reformatted - it happens for any reflow or text block reformatting

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -114,20 +119,59 @@ private static ImmutableSet<Range<Integer>> rangesAfterAppliedReplacements(
     }
 
     @SuppressWarnings("for-rollout:NullAway")
-    private static TreeRangeMap<Integer, String> getReflowReplacements(int columnLimit, final String input)
+    public static TreeRangeMap<Integer, String> getReflowReplacements(int columnLimit, final String input)
             throws FormatterException {
-        JCTree.JCCompilationUnit unit = parse(input, /* allowStringFolding= */ false);
-        String separator = Newlines.guessLineSeparator(input);
+        return new Reflower(columnLimit, input).getReflowReplacements();
```
crogoz, line 128: tried to keep this in line with pjf logic - refactored to be similar to pjf - as it easier to compare diffs.

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +185,112 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+                String stripped = initialLines.stream()
+                        .skip(1)
+                        .collect(joining(separator))
+                        .stripIndent();
+                ImmutableList<String> lines = stripped.lines().collect(ImmutableList.toImmutableList());
+
+                String prefix = (lineStartPosition + startColumn + 4 > startPosition ? "" : " ".repeat(4))
+                        + " ".repeat(startColumn);
+
+                StringBuilder output = new StringBuilder(initialLines.get(0));
```
crogoz: diverges from gjf to ensure we indent text blocks

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -381,13 +480,56 @@ private static int getStartPosition(Tree tree) {
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns true if the lines containing the {@code initialRangesToChange} in the given Java source exceed the
+     * column limit or contain text blocks. Used by the Intellij plugin to check if we need to run StringWrapper.wrap.
+     * Keep this method and {@code needWrapping} in line.
+     * */
+    public static boolean linesNeedWrapping(int columnLimit, String input, RangeSet<Integer> initialRangesToChange) {
```
crogoz, line 607: pjf specific - we need to figure out if we need to wrap the current input in the rangeSet initialRangesToChange

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.output
```diff
@@ -1,6 +1,106 @@
...
+            foo
+            """ + """
+        bar
+        """;
+        String working = """
+            foo
+            """ + """
+            bar
+            """;
+        String u = stringVariableOne + """
```
crogoz: unfortunately string concatenation is not great with text blocks and I am not sure if I can change it. 
after running the AST formatter I am getting (The AST cannot change the indentation of the text block node): 
```
    String u = stringVariableOne + """
            ...
            """ + stringVariableTwo + """
                    ...
                    """;
```
and then after I try to re-indent the text blocks and given that I am trying to indent based on the previous line, I am getting the block below. 

For these kind of text block operations, people should try to keep the textBlocks indented at the same line with the rest of the operands and then the block will be properly indented

crogoz:
Depends on https://github.com/palantir/gradle-baseline/pull/3276. Adding `do not merge` until the previous PR is merged

message:
Waiting

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +185,112 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
-            int fistLineCol = lineMap.getColumnNumber(getStartPosition(startingPath.getLeaf())) - 1;
-
-            // Handling leaving trailing non-string tokens at the end of the literal,
-            // e.g. the trailing `);` in `foo("...");`.
-            @SuppressWarnings("for-rollout:NullAway")
-            int end = getEndPosition(unit, getLast(flat));
-            int lineEnd = end;
-            while (Newlines.hasNewlineAt(input, lineEnd) == -1) {
-                lineEnd++;
+        private void indentTextBlocks(TreeRangeMap<Integer, String> replacements, List<Tree> textBlocks) {
```
CRogers: In general the changes in this PR look much better! I ran it on [gradle-guide here](https://github.com/palantir/gradle-guide/commit/e0d9657d8b83b2984f0b718581c100bd17481794) and most of the changes are an improvement imo.

The only weirdness (on this repo at least) is around multiple text block arguments to the same method, example:

```java
refactorFromTo("""
    someCode
    """, """
        otherCode
        """);
```

I think the second text block is getting indented out of line with the first because the first ending `"""` has been indented?

You can take this to it's logical extreme:

```java
void lotsOfStringParams(String a, String b, String c, String d, String e, String f, String g) {}

void test() {
    lotsOfStringParams("""
        aaa
        """, """
            bbb
            """, """
                ccc
                """, """
                    ddd
                    """, """
                        eee
                        """, """
                            fff
                            """, """
                                ggg
                                """);
}
```

CRogers: Ah, this might be the same as a concatenation issue?

CRogers: If you force it to put each arg on a new line by making one arg very long, it kinda looks better but stops the `""", """` pattern:

```java
    void lotsOfStringParams(String a, String aa, String b, String c) {}

    void test() {
        lotsOfStringParams(
                """
                aaa
                """,
                "sdklfjslfkjsadlkgjsdklfjsadlkfjsaklfjaskdsfkljsaklfjsadflkjsdafkljasdfklsjfklsajflkasjfsfjksflfj",
                """
                bbb
                """,
                """
                ccc
                """);
    }
```

crogoz: re: last example with the one long arg
I believe this has to do with how parameters for a method are going to be structured if the maxLineColumn would exceed - same behaviour if we had long variable names/long annotated variables. 

```
      void test(
              String aaaaaaaaaa,
              @NotNull @Variable Ssdklfjslfkjsdasdadasasdsadsadasdasadadasddadsasdsadadasdadadasdas variable,
              String bbbbbb) {
          lotsOfStringParams(
                  aaaaaaaaaa,
                  sdklfjslfkjsdasdadasasdsadsadasdasadadasddadsasdsadadasdadadasdasadadasdasfasfasfasfas,
                  bbbbbb);
      }
```

crogoz:
adding `do not merge` as I am trying a workaround for the concatenation & method invocation extra indent issue from: https://github.com/palantir/palantir-java-format/pull/1419#discussion_r2394538870

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.output
crogoz: this is not valid anymore - should be fixed by: https://github.com/palantir/palantir-java-format/pull/1419/commits/ab9282c2b065676f7c70d6a728f79523b84e5fc7

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.output
```diff
@@ -1,6 +1,155 @@
...
+    String w = lotsOfStringParams(
+            """
+            aaa
+            """, "sdklfjslfkjsadlkgjsdklfjsadlkfjsaklfjaskdsfkljsaklfjsadflkjsdafkljasdfklsjfklsajflkasjfsf", """
+            bbb
+            """, """
+            ccc
+            """);
+
+    String w = lotsOfStringParams(
```
crogoz, line 144: new line for each parameter - this has to do with the default way we are breaking in case some params are long (https://github.com/palantir/palantir-java-format/pull/1419/files#diff-1525009326c7709801fbaf30e5963924f4ab6e68f3d9ba1ebd3cc0b8e4c662c2R148).
I think this makes sense as is rn, the 3rd parameter wouldn't fit in any case on line 148, so for consistency I think this looks better than trying to move some of the initial quotes on the previous line.

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.output
```diff
@@ -1,6 +1,155 @@
...
+        ddd
+        """, """
+        eee
+        """, """
+        fff
+        """, """
+        ggg
+        """);
+
+    String w = lotsOfStringParams(
```
crogoz, line 135: Ideally the initial quotes would be on the previous line; but because of how params are split to fit into the 120 limit, we are adding the initial paramater on a new line - same behaviour below with a normal variable:

```
    String w = lotsOfStringParams(
            a, "sdklfjslfkjsadlkgjsdklfjsadlkfjsaklfjaskdsfkljsaklfjsadflkjsdafkljasdfklsjfklsajflkasjfsf", """
            bbb
            """, """
            ccc
            """);
```

crogoz:
In gjf the implementation is quite straightforward for text blocks - it makes sure the ending """ has the same indentation as the starting """. 

In our case, because the """ can be either:
1.  on the same line as the parent => we will try to indent the text block 4 spaces - see below:
```
String value = """
 my text block here // we need to indent this by 4 spaces - for readability
 """
```
2. when a break happens (eg. line would be too long to fit multiple params- like below -  the default indentation in the AST nodes is 8):
```
String value = value(
       """
   my text block here // we need to indent these 2 lines by 8 spaces (enforced by the AST break, see initial """ or the param below)
   """,
       " my extremely looooooooooooooooooooooooooooooooooooooooooooooooooooong value that triggers the break",
      .....);
```

I implemented a heuristic to make sure the indentation of the textBlocks is consistent (logic [here](https://github.com/palantir/palantir-java-format/pull/1419/commits/8169884ac67a8df6eb94a5e357e9b7ad1d107363)), where I special case the concatenation (PLUS trees eg. `'textBlock' + 'other'`) and method params ( `method(textBlockParam, other, ....)`)
For each textBlock:
1. search the enclosing parent & figure out the maximum indentation level for it's kids (enforced by the AST code)
2. store the mapping between the textBlock and the desired indentation level

For the basic cases, I think my logic covers them but for more complicated scenarios, we'd get something like this: 
Screenshot 2025-10-08 at 5 16 46 PM
Screenshot 2025-10-08 at 5 14 55 PM

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -2656,6 +2655,7 @@ void visitDot(ExpressionTree node0) {
         List<ExpressionTree> items = new ArrayList<>(stack);
 
         boolean needDot = false;
+        boolean isTextBlock = false;
```
CRogers: Can we add this variable inside the `else` starting on 2669? We should try to keep it's lifetime as bounded as possible.

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+                    }
+                    if (i == lines.size() - 1) {
+                        String withoutDelimiter = line.substring(0, line.length() - TEXT_BLOCK_DELIMITER.length())
+                                .stripTrailing();
+                        if (!withoutDelimiter.isEmpty()) {
+                            output.append(withoutDelimiter)
+                                    .append('\\')
+                                    .append(separator)
+                                    .append(prefix);
+                        }
```
CRogers, line 243: What is this bit doing?

palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java
CRogers: Am I reading around this code correctly that this `FormatFileCallable` is only just by the cli `Main` class and the existing `FormattingService` in the intellij plugin called `FormatterServiceImpl` which does reflow strings? So the bug where the strings are not reflowing in intellij is only when using the native formatter (and so the cli)?

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+            int startParentLine = lineMap.getLineNumber(startParentPosition);
+            int endParentPosition = getEndPosition(unit, parentPath);
+            int endParentLine = lineMap.getLineNumber(endParentPosition);
+            int lineParentStartPosition =
+                    lineMap.getStartPosition(shouldUseStartLineParent ? startParentLine : endParentLine);
+            int startParentColumn = CharMatcher.whitespace()
+                    .negate()
+                    .indexIn(input.substring(lineParentStartPosition, endParentPosition));
+            int extraIndent = 4;
+            RangeSet<Integer> allRanges = TreeRangeSet.create();
```
CRogers: It seems like we just add to this `RangeSet` but never read from it? Can we remove this?

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
CRogers: Appreciate it trying to keep it similar to GJF. I think the only big string syntax change is going to be [String Templates](https://openjdk.org/jeps/465) - hopefully we can reuse basically the same logic, it looks to be the same as text blocks with a `STR."""` or similar at the start instead.

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+
+        /**
+         * Pjf specific: When the AST needs to break a concatenation expression ('ONE'+'TWO') or the parameters of a method,
+         * the indentation will be set to 8, in which case we need to align the text blocks to the correct indentation.
+         * In order to compute the indentation value we need to do:
+         * 1. for each textBlock find the enclosing block/parent (only for concat expressions/method invocations)
+         * 2. for each parent, find the arguments/concatenated expressions and find the max indentation level
+         * (ignore the lines that start with a text block ending)
+         * 3. store the mapping between the textBlock and the indentation level computed before.
+         */
```
CRogers, line 264: Great comment, I would be so lost without it

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+            for (TreePath textBlock : textBlocks) {
+                TreePath parentPath = textBlock.getParentPath();
+                Tree parent = parentPath.getLeaf();
+                if (parent instanceof MethodInvocationTree) {
+                    textBlockToParent.put(textBlock, parent);
+                    if (parentToIndent.containsKey(parent)) {
+                        continue;
+                    }
+                    List<Tree> allArguments = new ArrayList<>(((JCMethodInvocation) parent).getArguments());
+                    parentToIndent.put(
```
CRogers: `parentToIndent.computeIfAbsent`? Rather than if/continue?

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+                } else if (parent.getKind() == Kind.PLUS) {
+                    while (parentPath.getParentPath().getLeaf().getKind() == Kind.PLUS) {
+                        parentPath = parentPath.getParentPath();
+                    }
+                    parent = parentPath.getLeaf();
+                    textBlockToParent.put(textBlock, parent);
+                    if (parentToIndent.containsKey(parent)) {
+                        continue;
+                    }
+                    parentToIndent.put(parent, computePrefixIndentation(parent, flattenExpressionTree(parent), true));
```
CRogers: Similarly I think we could leverage `computeIfAbsent` here.

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+                            //      .method("arguments")
+                            //      .build()
+                            // In this case the parent of the arguments should be the method ("Class.builder().method")
+                            // the indentation of the arguments will be relative to the indentation of the last line of
+                            // the method name.
+                            computePrefixIndentation(
+                                    ((JCMethodInvocation) parent).getMethodSelect(), allArguments, false));
+                } else if (parent.getKind() == Kind.PLUS) {
+                    while (parentPath.getParentPath().getLeaf().getKind() == Kind.PLUS) {
+                        parentPath = parentPath.getParentPath();
```
CRogers, line 287: I think the usage of the term `parent` is a little overloaded here - we're not actually interested in the direct `parent` as it's already named in the compiler, but the "logical parent" aka "the start of the method arg expression" or "beginning of concatenation"? I can't really think of a good name here - but I think anything to differentiate it from the regular treepath parent would be good here

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -114,20 +123,59 @@ private static ImmutableSet<Range<Integer>> rangesAfterAppliedReplacements(
...
-        List<TreePath> toFix = new ArrayList<>();
-        final Position.LineMap lineMap = unit.getLineMap();
-        new TreePathScanner<Void, Void>() {
             @Override
             public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
                 if (literalTree.getKind() != Kind.STRING_LITERAL) {
                     return null;
                 }
+                int pos = getStartPosition(literalTree);
+                if (input.substring(pos, Math.min(input.length(), pos + 3)).equals(TEXT_BLOCK_DELIMITER)) {
```
CRogers: `TEXT_BLOCK_DELIMITER.length()` instead of hardcoded `3` - just to make it more obvious what `3` refers to?

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.output
```diff
@@ -1,6 +1,185 @@
...
+            """
+            aaa
+            """,
+            "sdklfjslfkjsadlkgjsdklfjsadlkfjsaklfjaskdsfkljsaklfjsadflkjsdafkljasdfklsjfklsajflkasjfsfdasdsadasdsada",
+            """
+            bbb
+            """,
+            """
+            ccc
+            """);
```
CRogers, line 154: This is all looking so much better! Such an improvement over the old formatting for strings!

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
```diff
@@ -141,59 +189,206 @@ public Void visitLiteral(LiteralTree literalTree, Void aVoid) {
...
+                allRanges.add(Range.closed(startLine, endLine));
+                int parentLine = shouldUseStartLineParent ? startParentLine : endParentLine;
+                // ignore indentation if the current argument is on the same line as the parent
+                if (startLine == parentLine) {
+                    continue;
+                }
+                // if this is a line that ends a textBlock (if the line starts with triple quotes & the current tree
+                // starts after)
+                int startColumn = CharMatcher.whitespace().negate().indexIn(input.substring(lineStartPosition, endPos));
+                if (input.startsWith("\"\"\"", lineStartPosition + startColumn)
```
CRogers: Can we use `TEXT_BLOCK_DELIMITER`?

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
crogoz: ups, yes - from a previous heuristic I tried 😅  I wonder why it doesn't show up as unused as I am never reading it

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
crogoz: we always want the textBlock to finish with the ending `""" `on a new line - so in case the input looks like [this](https://github.com/palantir/palantir-java-format/pull/1419/files#diff-e572e4373d8fd1705e1ec67ee3dce54319a23b996d0321b9c5a3e1e8b70a8d16R3-R6) - notice that the last line ends with the last line of the text block and the ending quotes on the same line, we will transform it to [this](https://github.com/palantir/palantir-java-format/pull/1419/files#diff-2bf37a530dfce4a30ab46929000865343720f30163cd261d8dd076bab563a083R5-R6), where the ending quotes will be on a new line  and the textBlock will finish with `\` - such that it has the same semantic meaning

palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java
crogoz: Hmm, no, I think this is always the case when running through the Intellij plugin (same bug that gjf has here as well issue 566 in gjf)

When running the [Intellij plugin we will always try to run the getFormatReplacements](https://github.com/palantir/palantir-java-format/blob/develop/idea-plugin/src/main/java/com/palantir/javaformat/intellij/PalantirJavaFormatFormattingService.java#L104-L105) which doesn't include the String.wrapper call (I guess this is because the String.wrapper will always reformat the whole file, whereas in the intellij plugin we might get a subset of lines from the initial file to be formatted). However, through cli, we are calling the other [mehtod](https://github.com/palantir/palantir-java-format/blob/1cc116070dcbe74c14b5c870e66d395a6d0d9a71/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/spotless/PalantirJavaFormatStep.java#L72) which always calls the StringWrapper.

palantir-java-format/src/main/java/com/palantir/javaformat/java/FormatFileCallable.java
```diff
@@ -49,14 +50,29 @@ public String call() throws FormatterException {
...
+                // if no wrapping change happened, return the initial replacements
+                if (wrappedFormattedText.equals(formattedText)) {
+                    return writeFormatReplacements(replacements);
+                }
+                return writeFormatReplacements(List.of(Replacement.create(0, input.length(), wrappedFormattedText)));
```
crogoz, line 67: worst case scenario here - when string reflows happens/text blocks are indented, then we will return the full range of the file to be replaced. This could be optimized - I think there is a way to check where the text was reformatted and return only those ranges. I can add this as a FLUP to the PR.

palantir-java-format/src/main/java/com/palantir/javaformat/java/StringWrapper.java
CRogers: I don't think intellij is smart enough to know that's it's not being read. There could be side effects from calling `.add` for all it knows the rest of the code could be relying on.

CRogers (approved):

talios:
I see this is now merged/released, I just updated to the latest release (about 2 beyond this) and removed my `// spotless:off` comments around text blocks, which were indented about 10 characters, running `spotless:apply` via Apache Maven I see the text blocks still become flush at character 0.

What was:

```java
    var token = buildToken("""
      {
        "token_type": "Bearer", .....
```

became

```java
    var token = buildToken(
        """
{
"token_type": "Bearer",.....
```

Which doesn't look right. Was this issue intended to be addressed at all, or is further work still needed on text block support?

crogoz:
Text block formatting should be fully supported now (palantir-java-format >= 2.78.0)

@talios could you please open a new github issue for this and add the full example of the invalid formatted text block ? You can also try the repro the issue by adding the text block in [here](https://github.com/palantir/palantir-java-format/blob/develop/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/RSL.input) and running the [tests](https://github.com/palantir/palantir-java-format/blob/develop/palantir-java-format/src/test/java/com/palantir/javaformat/java/FormatterIntegrationTest.java#L72). Thanks!

PR #1420, opened 2025-09-25
Fix comments after default switch case expressions

crogoz:
## After this PR

Fixes parsing of switch expressions that have a comment before the closing "}". (internal example [here](https://pl.ntr/2x3) with the corresponding failure [here](https://pl.ntr/2x4)). Internal discussion [here](https://pl.ntr/2x5)
==COMMIT_MSG==
Fix comments after default switch case expressions
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.input
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.input
@@ -58,4 +58,17 @@ class ExpressionSwitch {
           case 2 -> throw new IllegalArgumentException();
           default -> throw new IllegalStateException();};
   }
+
+
+    public void testWithComments(int y) {
+        int x;
+        x = switch (y) {
+            case 1 -> 1;
+            // after case 1
+            case 2 -> throw new IllegalArgumentException();
+            // after case 2
+            default -> throw new IllegalStateException();
+            // after default case
+        };
+    }
 }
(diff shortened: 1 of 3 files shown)


esword (dismissed):

palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/ExpressionSwitch.input
```diff
@@ -58,4 +58,17 @@ class ExpressionSwitch {
           case 2 -> throw new IllegalArgumentException();
           default -> throw new IllegalStateException();};
   }
+
+
+    public void testWithComments(int y) {
```
felixdesouza, line 63: is this the intended snippet for the `before`? I couldn't see what the difference was manually, and both diff and difft report no differences.

crogoz: yes, without this change, the formatter would put the "}" on the same line as the comment - which breaks the compilation eg. 
```
        x = switch (y) {
            case 1 -> 1;
            // after case 1
            case 2 -> throw new IllegalArgumentException();
            // after case 2
            default -> throw new IllegalStateException();
            // after default case};
```

felixdesouza: ah, are you able to put that in the description, it's quite hard to tell what's being fixed from just the code + PR description

crogoz: I've changed the input to a wrongly formatted switch expression [here](https://github.com/palantir/palantir-java-format/pull/1420/commits/78dc77067779470ce601187fe323b617c8d1c7b9)

felixdesouza:
👍  I keep forgetting invalidated on push on public gh

PR #1422, opened 2025-09-29
Remove .github/workflows/gradle-wrapper-validation.yml

aldexis:
## Before this PR
This was added to a few repos by excavator, but isn't something we necessarily want anymore.

~Note that this specific one was added by @fawind in https://github.com/palantir/palantir-java-format/pull/887 though, so maybe we want to keep it?~
Edit: that PR only edited it - it was added in https://github.com/palantir/palantir-java-format/pull/185

## After this PR

==COMMIT_MSG==
Remove .github/workflows/gradle-wrapper-validation.yml
==COMMIT_MSG==


CRogers:
:+1:

PR #1428, opened 2025-10-06
Fix latest Intellij incompatibility

crogoz:
## Before this PR

Screenshot 2025-10-06 at 10 33 31 AM

## After this PR
Remove unused method that is also causing incompatibility with Intellij version: 2025.3 eap (253.22441.33)

==COMMIT_MSG==
Fix latest Intellij incompatibility
==COMMIT_MSG==


FinlayRJW:
👍

PR #1433, opened 2025-10-13
Java files always end with a new line

crogoz:
## Before this PR

Checkstyle requires Java files to  end with a new line see [here](https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java-config/resources/checkstyle/checkstyle.xml#L18-L20).

## After this PR

We are now enforcing this through the formatter directly.
Fixed a bug where Bootstrapping/NativeImageFormatter were incorrectly reading the input range as a 1-index vs 0-index. 

==COMMIT_MSG==
Java files always end with a new line
==COMMIT_MSG==


palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
```diff
@@ -384,6 +384,8 @@ public Void visitCompilationUnit(CompilationUnitTree node, Void unused) {
         handleModule(first, node);
         // set a partial format marker at EOF to make sure we can format the entire file
         markForPartialFormat();
+        // Ensure the file ends with a newline
+        builder.forcedBreak();
```
CRogers, line 388: Can we write a test for this?

CRogers: Also perhaps the multiple lines at the end of the file case

crogoz:
Adding `do not merge`, the Intellij plugin seems to ignore the last lines when returing the list of replacements. Investigating now why that is the case.

palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptionsParser.java
```diff
@@ -79,7 +79,7 @@ static CommandLineOptions parse(Iterable<String> options) {
                 case "-character-ranges":
                 case "--character-range":
                 case "-character-range":
-                    parseRangeSet(optionsBuilder.characterRangesBuilder(), getValue(flag, it, value));
+                    parseCharacterRanges(optionsBuilder.characterRangesBuilder(), getValue(flag, it, value));
```
crogoz, line 82: Bug from the beginning of pjf. `parseRangeSet` expects a 1-indexed range, but we have a 0-indexed range (see RnageUtils class).

This would mean that Intellij only (./gradlew spotless would work correclty as it doesn't use the same `getReplacements` workflow) would not reformat the full file eg. after formatting the "format.input" file, we'd get a truncated version of the formatted output: 

Screenshot 2025-10-14 at 5 30 11 PM

palantir-java-format-jdk-bootstrap/src/test/resources/format.input
```diff
@@ -9,4 +9,4 @@ class B {
         break;
     }
   }
-}
+}
```
crogoz, line 12: intentionally removing the last line to make sure we are getting the error seen in: https://github.com/palantir/palantir-java-format/pull/1433/files#r2429564482

palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptionsParser.java
CRogers: Nice find!

CRogers (approved):
This is great, should remove this toil for devs. I checked as well and intellij will format after making a new file, so this is immediately correct.

PR #1055, opened 2024-03-18
Add shadowJar for generating all-deps jar

kyu-sz:
## Before this PR

There is no all-deps jar created. People raise issues like this: https://github.com/palantir/palantir-java-format/issues/902

## After this PR

We can run `./gradlew shadowJar` to generate all-deps jar.

## Possible downsides?

None

PR #1464, opened 2025-11-06
improvement: Publish executable shadowJar for formatter with CLI

andreweacott:
Configure the palantir-java-format module to publish the shadowJar (fat JAR with all dependencies) as an additional artifact with 'executable' classifier. This enables distributing a self-contained formatter jar alongside the standard library artifacts.

## Before this PR
The palantir-java-format Jar artifacts published from CI are not executable, making it hard to use the formatter as a CLI

## After this PR
* A shadow Jar is built alongside the default formatter Jar
* The shadow Jar is additionally published to Maven with the 'executable' classifier

## Possible downsides?
None I can think of


norbertwnuk:
There are scenarios where JAR is preferred over native artifact (mentioned in the main README.md), from my perspective it would be easier to have drop-in replacement for https://github.com/NixOS/nixpkgs/blob/nixos-26.05/pkgs/by-name/go/google-java-format/package.nix#L42. As such interested in this PR even when the native build is published already: https://mvnrepository.com/artifact/com.palantir.javaformat/palantir-java-format-native 

Replaces: https://github.com/palantir/palantir-java-format/pull/1055

PR #1483, opened 2025-11-25
get rid of truth

esword:
## Before this PR
This repo has been failing Upgrade Dependencies excavator for almost 2 years (since Feb 29, 2024). https://github.com/palantir/palantir-java-format/pulls?q=is%3Apr+%22Excavator%3A+Upgrade+dependencies%22+ .  It's gone through a series of PRs that it opened and closed in that time.  This is the current one - https://github.com/palantir/palantir-java-format/pull/1473

The reason (I think) is that it uses a library called com.google.truth:truth for an assertThat and there are (possibly) multiple versions of it.  Why we use this library _and_ also use assertj, I don't know.  I've removed uses of truth to hopefully unblock other things.

And yes, I did get dark amusement out of the subject of this PR.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/build.gradle
+++ b/palantir-java-format/build.gradle
@@ -15,7 +15,6 @@ dependencies {
     implementation 'com.fasterxml.jackson.module:jackson-module-parameter-names'
 
     testImplementation 'com.google.guava:guava-testlib'
-    testImplementation 'com.google.truth:truth'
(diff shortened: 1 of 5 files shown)


felixdesouza:
👍  I believe truth is coming from google-java-format which this repo is forked from, it's their own assertion library iirc. Don't see anything wrong with getting rid of truth here

PR #1490, opened 2025-12-01
Improve logging when `git diff` fails

kelvinou01:
## Before this PR

When `git diff` fails, we don't tell the user the exit code and stderr/stdout, meaning the user has to dive in and debug.

## After this PR

==COMMIT_MSG==
Improve logging when `git diff` fails
==COMMIT_MSG==


gradle/.DS_Store
esword, line 1: How did this get added?

kelvinou01: yikes, no idea!

esword (dismissed):

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/FormatDiff.java
```diff
@@ -134,8 +134,11 @@ private static String gitCommand(Path dir, String... args) throws IOException, I
                 new ProcessBuilder().command(args).directory(dir.toFile()).start();
 
         Preconditions.checkState(process.waitFor(30, TimeUnit.SECONDS), "git diff took too long to terminate");
-        Preconditions.checkState(process.exitValue() == 0, "Expected return code of 0");
-
+        if (process.exitValue() != 0) {
+            throw new IllegalStateException("Expected return code of 0 but got " + process.exitValue()
```
crogoz: I'd use a `RuntimeException` here instead, IllegalStateException is more to signal an[ illegal state](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalStateException.html) of the application

kelvinou01: Gotcha, i used illegalstate to maintain parity with the previous exception checkState throws, but a runtime exception does make more sense here

felixdesouza:
👍

PR #1521, opened 2026-01-06
Fix Gradle 9 compatibility for XML manipulation tasks

gavinbunney:
## Before this PR
When upgrading to Gradle 9, the `updatePalantirJavaFormatXml` and `updateWorkspaceXml` tasks fail with `NoClassDefFoundError: groovy/util/XmlParser`. 

Gradle 9 no longer includes Groovy classes on the runtime classpath, and attempting to add explicit Groovy 4 dependencies creates version conflicts with Gradle 8's bundled Groovy 3. This blocks the upgrade to Gradle 9.

## After this PR
Both Gradle v8 and Gradle v9 are supported via switching out the use of Groovy's XML APIs (`XmlParser`/`XmlNodePrinter`) with Java's standard DOM APIs (`javax.xml.parsers`, `javax.xml.transform`).

There is a fair bit of code added to adopt between the groovy `Node` API the std java DOM types; had tried using the Java DOM APIs in the calling `ConfigureJavaFormatterXml` directly instead, to remove the need for this adoption layer, but it ended up being harder to read from the callers side.

==COMMIT_MSG==
Add support for Gradle v9 via removing dependency on Groovy 3 xml parsers
==COMMIT_MSG==

### DIFF
--- a/gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/XmlUtils.java
+++ b/gradle-palantir-java-format/src/main/groovy/com/palantir/javaformat/gradle/XmlUtils.java
@@ -16,44 +16,131 @@
 
 package com.palantir.javaformat.gradle;
 
-import com.google.common.collect.ImmutableMap;
-import com.google.common.io.Files;
 import groovy.util.Node;
-import groovy.util.XmlNodePrinter;
-import groovy.util.XmlParser;
-import java.io.BufferedWriter;
 import java.io.File;
-import java.io.IOException;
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @gavinbunney! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1531, opened 2026-01-13
Spotless excludes  formatting of `**/build/*` files

crogoz:
## Before this PR

We were formatting [all javaSource files](https://github.com/diffplug/spotless/blob/main/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/JvmLang.java#L38), even if the files are generated  (most of the times they end up being in the build/ directory and not committed).

## After this PR
Perf improvement for internal repo from https://pl.ntr/2Dq (4m 30s) -> https://pl.ntr/2Do (3m 56.530s)
==COMMIT_MSG==
Spotless excludes  formatting of `**/build/*` files
==COMMIT_MSG==


CRogers:
:+1:

PR #1568, opened 2026-02-10
Exclude generated directories for formatting

crogoz:
## Before this PR

FLUP for https://github.com/palantir/palantir-java-format/pull/1531
Some source directories might have `build` in the directory structure; this regex  would ignore these files. 

## After this PR
Exclude generated directories - same logic as in error-prone [here](https://github.com/palantir/suppressible-error-prone/blob/611ff7acd4462afd6aa0464ea6925c0190bd556f/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L217-L224). Internal thread [here](https://pl.ntr/2Eh).

==COMMIT_MSG==
Exclude generated directories
==COMMIT_MSG==

### DIFF
--- a/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
+++ b/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
@@ -51,7 +51,10 @@ public abstract class SpotlessInterop implements Action<JavaExtension> {
 
     @Override
     public void execute(JavaExtension java) {
+        // Exclude build directories at root and in subprojects
         java.targetExclude("**/build/**/*");
+        // Re-include any build directories that are inside src directories
+        java.target("**/src/**/build/**/*");
         // This is configuration cache safe as happening afterEvaluate
         java.addStep(spotlessJavaFormatStep());
     }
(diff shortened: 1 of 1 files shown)


gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
```diff
@@ -51,7 +51,11 @@ public SpotlessInterop(Supplier<FormatterService> formatterService) {
 
     @Override
     public void execute(JavaExtension java) {
-        java.targetExclude("**/build/**/*");
+        // Exclude generated source directories
+        java.targetExclude("**/build/generated*/**");
+        java.targetExclude("**/src/generated*/**");
+        java.targetExclude("**/generated_*src/**");
+        java.targetExclude("**/generated_*Src/**");
```
kelvinou01, line 60: might be useful to explain here that some projects have `build` as a folder name, so we can't blanket ignore it

kelvinou01 (commented):
LGTM other than small nit

kelvinou01 (approved):

PR #1579, opened 2026-02-15
Fix parsing of unnamed patterns

mrjameshamilton:
## Before this PR

I came across FormatException of the form `error: expected token: '_'; generated ) instead` when using unnamed patterns, like the following (from the JEP):

```
class UnnamedPattern {
    public static void main(String[] args) {
        var cp = new ColoredPoint(new Point(1, 2), Color.RED);

        // The colour doesn't matter, so use an unnamed pattern
        // (as opposed to the unnamed type pattern `var _`).
        if (cp instanceof ColoredPoint(Point(int x, int y), _)) {
            System.out.println(x + "," + y);
        }

        if (cp instanceof ColoredPoint(_, var c)) {
            System.out.println(c);
        }
    }
}

record Point(int x, int y) { }
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) { }
```

I resolved this by overriding `visitAnyPattern(AnyPatternTree node, Void unused)` in the existing `Java21InputAstVisitor`.

However, `AnyPatternTree` is annotated as a preview feature in Java 21 so I changed Java versions from 21 to 23.

Before changing the version, I tried adding `--enable-preview` but then I got an error message that `--enable-preview` is not compatible with Java 11 source.

I also noticed Java 23 mentioned in the root `build.gradle` so I changed it to 23 rather than 22, 24, 25.

One thing I found in this change was that `getImports()` now returns a `JCImportBase`, so to ensure backwards compatibility I added an `instanceof` check that (and module imports support would need to be implemented in the future).

## After this PR
==COMMIT_MSG==
Fix parsing of unnamed patterns
==COMMIT_MSG==

## Possible downsides?

I'm not sure if changing Java versions from 21 to 23 (runtime in root build.gradle, target in `palantir-java-format-spi/build.gradle` and target in `palantir-java-format/build.gradle`) will have any negative effect? And if another solution should be found to resolve the issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -75,7 +75,7 @@ subprojects {
 
 javaVersions {
     libraryTarget = 17
-    runtime = 21
+    runtime = 23
 }
 
 jdks {
--- a/palantir-java-format-spi/build.gradle
+++ b/palantir-java-format-spi/build.gradle
@@ -18,5 +18,5 @@ tasks.withType(JavaCompile).named('compileJava') {
 }
 
 javaVersion {
-    target = 21
+    target = 23
 }
--- a/palantir-java-format/build.gradle
+++ b/palantir-java-format/build.gradle
@@ -41,6 +41,16 @@ moduleJvmArgs {
             'jdk.compiler/com.sun.tools.javac.api')
 }
 
+applicationDefaultJvmArgs = [
+        '--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED',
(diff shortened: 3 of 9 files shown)


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @mrjameshamilton! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

almarzn:
Is this MR planned to be merged?

oualidbklee:
is this PR planned to be merged ?

PR #1630, opened 2026-03-24
 Fix crash when formatting method references 

crogoz:
## Before this PR

Internal thread [here](https://pl.ntr/2Al)
Formatting code containing a method reference whose qualifier is a simple method call (e.g. `getModificationContext()::getRestrictionStatusOrDefault`) threw:
```
<File>.java:LINE_UNDEFINED palantir-java-format(java.lang.IllegalStateException) Didn't find expected break at the beginning of level. (...)
  Resolve these lints or suppress with `suppressLintsFor`
```

This happened because `CountWidthUntilBreakVisitor.visitLevel` — which computes how wide a prefix is before its first break — assumed all levels either had a special inlineability hint or started with nested Level objects. A simple method call starts with a `Token`, which matched neither case and fell through to a throw.

## After this PR

The fallback instead walks the level's docs, summing their widths until the first Break.

==COMMIT_MSG==
 Fix crash when formatting method references 
==COMMIT_MSG==

### DIFF
--- a/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/A.output
+++ b/palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/A.output
@@ -70,5 +70,18 @@ class A {
         something = 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2
                 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2
                 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2;
+        return StreamEx.of(itemsToCreate)
+                .append(itemsToEdit)
+                .append(itemsToDelete)
+                .distinct()
+                .mapToEntry(item -> getArgsGenerator()
+                        .getNewArgs(
+                                extendedItems.getOrDefault(item, Set.of()),
+                                getOwningResource(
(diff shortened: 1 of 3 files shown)


palantir-java-format/src/test/resources/com/palantir/javaformat/java/testdata/A.output
```diff
@@ -70,5 +70,18 @@ class A {
         something = 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2
                 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2
                 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2 + 2;
+        return StreamEx.of(itemsToCreate)
```
felixdesouza, line 73: I don't understand this diff

crogoz: this in the input, the output is in A.output, previously it was failing at [this line](https://github.com/palantir/palantir-java-format/pull/1630/changes#diff-2d628df3c3a24b95156517fbb14ebfbe9b7a05ccdb2565e7f249348d506bbe16R92). The tests are making sure that formatting A.input will result in A.output

felixdesouza: this is `A.output` no? I suppose I didn't understand the diff because it didn't look like there was anything different? But you're basically saying the equivalent of: `assertThatCode(...).doesNotThrow()`?

crogoz: yes, exactly!

felixdesouza (approved):

PR #1672, opened 2026-05-20
Fix markdown docstring wrapping

DHerls:
## Before this PR
When there was a very long markdown docstring comment (beginning with `///`), the formatter would break the comment up, but would only add `//` to the beginning of the new line instead of `///`.

```java
/// Very long markdown docstring with an arbitrary cutoff
class MyClass {}
``` 

```java
/// Very long markdown docstring with an arbitrary
// cutoff
class MyClass {}
``` 

## After this PR
==COMMIT_MSG==
Fix issue when wrapping markdown docstrings resulted in 2 slashes instead of 3
==COMMIT_MSG==

## Possible downsides?
This fix piggybacks on the regular comment line formatting instead of teaching the formatter about markdown docstring comments. I think this is fine as a quick fix, but if any more problems come up with markdown docstring comments we should look into creating a dedicated formatter.

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaCommentsHelper.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaCommentsHelper.java
@@ -113,6 +113,14 @@ public final class JavaCommentsHelper implements CommentsHelper {
     private static final Pattern LINE_COMMENT_MISSING_SPACE_PREFIX =
             Pattern.compile("^(//+)(?!noinspection|\\$NON-NLS-\\d+\\$)[^\\s/]");
 
+    private static String lineCommentPrefix(String line) {
+        int prefixLength = 0;
+        while (prefixLength < line.length() && line.charAt(prefixLength) == '/') {
+            prefixLength++;
+        }
+        return "/".repeat(Math.max(prefixLength, 2));
+    }
+
     private List<String> wrapLineComments(List<String> lines, int column0) {
         List<String> result = new ArrayList<>();
         for (String line : lines) {
@@ -128,11 +136,7 @@ public final class JavaCommentsHelper implements CommentsHelper {
                 continue;
             }
             // Preserve the original slash prefix (e.g. `///` for markdown docstrings) on wrapped lines.
-            int prefixLength = 0;
-            while (prefixLength < line.length() && line.charAt(prefixLength) == '/') {
-                prefixLength++;
-            }
-            String prefix = "/".repeat(Math.max(prefixLength, 2));
+            String prefix = lineCommentPrefix(line);
             while (line.length() + column0 > options.maxLineLength()) {
                 int idx = options.maxLineLength() - column0;
                 // only break on whitespace characters, and ignore the leading `// `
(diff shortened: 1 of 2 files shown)


palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaCommentsHelper.java
```diff
@@ -113,6 +113,14 @@ private String indentLineComments(List<String> lines, int column0) {
     private static final Pattern LINE_COMMENT_MISSING_SPACE_PREFIX =
             Pattern.compile("^(//+)(?!noinspection|\\$NON-NLS-\\d+\\$)[^\\s/]");
 
+    private static String lineCommentPrefix(String line) {
+        int prefixLength = 0;
+        while (prefixLength < line.length() && line.charAt(prefixLength) == '/') {
+            prefixLength++;
+        }
+        return "/".repeat(Math.max(prefixLength, 2));
```
kelvinou01: i'm wondering why we have to set a minimum of 2? Is it possible for `line` to have no leading slashes?

kelvinou01 (commented):
also looping in someone else more familiar with PJF to take a look.

kelvinou01:
Can we do an `autorelease`? This is an external-facing behavior change.

palantir-java-format/src/test/java/com/palantir/javaformat/java/FormatterTest.java
```diff
@@ -430,6 +430,23 @@ public void blankLinesImportComment() throws FormatterException {
         assertThat(ImportOrderer.reorderImports(withoutBlank, Style.GOOGLE)).isEqualTo(withoutBlank);
     }
 
+    @Test
+    public void wrapMarkdownDocstringComment() throws Exception {
+        assertThat(Formatter.create()
```
kelvinou01: i know the rest of the file is poorly formatted, but can we use multiline strings for the test data? so we don't have to compute things in our head when trying to understand the test

kelvinou01: ~~we're on [target 17](https://github.com/palantir/palantir-java-format/blob/develop/build.gradle#L77C5-L77C23) so we should have multiline strings~~

nvm, i think we [still output java 11](https://github.com/palantir/palantir-java-format/commit/f59b03dba02095925b3029ebee6ab796481aab34#diff-b16b763408eeebb7ef338f52b978740f4ff49fce1a0186f730058ebcd527c30cR47-R50). nevertheless, can we extract the test data into variables so they're easier to read?

palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaCommentsHelper.java
DHerls: Yep good catch, Claude was a little over-cautious. We only get here if `tok.isSlashSlashComment()` is true.

palantir-java-format/src/test/java/com/palantir/javaformat/java/FormatterTest.java
DHerls: Pulled out into variables, let me know if this is better

crogoz (approved):

PR #1681, opened 2026-06-09
Only run the formatter on trusted projects

crogoz:
## After this PR

internal thread [here](https://pl.ntr/2C2)

==COMMIT_MSG==
Only run the formatter on trusted projects
==COMMIT_MSG==

Tested locally by running `./gradlew runIde`


FinlayRJW:
👍

PR #1690, opened 2026-06-15
Fix Spotless targetExclude patterns being overridden

crogoz:
## Before this PR

https://github.com/palantir/palantir-java-format/pull/1568 was not additive, calling `targetExclude `multiple times was overriding previous values instead of accumulating them.

## After this PR
FLUP for https://github.com/palantir/palantir-java-format/pull/1568
- Fix `SpotlessInterop` calling `targetExclude` multiple times, which overrides previous patterns instead of accumulating them - only the last exclude (`**/groovy-dsl-plugins/**`)  was actually applied
- Add `SpotlessExcludesTest` to verify `spotlessJavaCheck` skips files in all excluded  directories

==COMMIT_MSG==
Fix Spotless targetExclude patterns being overridden
==COMMIT_MSG==

### DIFF
--- a/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
+++ b/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
@@ -54,14 +54,15 @@ public abstract class SpotlessInterop implements Action<JavaExtension> {
         // Exclude generated source directories
         // Note: We cannot simply exclude **/build/**/* because some repos might contain a build directory
         // within the src sourceset that should be formatted.
-        java.targetExclude("**/build/generated*/**");
-        java.targetExclude("**/src/generated*/**");
-        java.targetExclude("**/generated_*src/**");
-        java.targetExclude("**/generated_*Src/**");
-        // build/groovy-dsl-plugins contains Java wrapper classes for
-        // https://docs.gradle.org/9.3.1/userguide/implementing_gradle_plugins_convention.html
-        // and don't need to be formatted.
-        java.targetExclude("**/groovy-dsl-plugins/**");
+        java.targetExclude(
+                "**/build/generated*/**",
+                "**/src/generated*/**",
+                "**/generated_*src/**",
+                "**/generated_*Src/**",
+                // build/groovy-dsl-plugins contains Java wrapper classes for
+                // https://docs.gradle.org/9.3.1/userguide/implementing_gradle_plugins_convention.html
+                // and don't need to be formatted.
+                "**/groovy-dsl-plugins/**");
         // This is configuration cache safe as happening afterEvaluate
         java.addStep(spotlessJavaFormatStep());
     }
--- a/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
+++ b/gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
@@ -52,9 +52,9 @@ public abstract class SpotlessInterop implements Action<JavaExtension> {
     @Override
     public void execute(JavaExtension java) {
         // Exclude generated source directories
-        // Note: We cannot simply exclude **/build/**/* because some repos might contain a build directory
-        // within the src sourceset that should be formatted.
         java.targetExclude(
+                // Note: We cannot simply exclude **/build/**/* because some repos might contain a build directory
+                // within the src sourceset that should be formatted.
                 "**/build/generated*/**",
                 "**/src/generated*/**",
                 "**/generated_*src/**",
--- a/gradle-palantir-java-format/src/test/java/com/palantir/javaformat/gradle/SpotlessExcludesTest.java
+++ b/gradle-palantir-java-format/src/test/java/com/palantir/javaformat/gradle/SpotlessExcludesTest.java
@@ -24,7 +24,6 @@ import com.palantir.gradle.testing.junit.GradlePluginTests;
 import com.palantir.gradle.testing.project.RootProject;
 import java.io.File;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
@@ -64,8 +63,9 @@ class SpotlessExcludesTest {
                 "build/generated/java",
                 "src/generated/java",
                 "src/main/generated_testsrc",
+                "src/main/generated_foo_testsrc",
                 "generated_testSrc",
-                "build/groovy-dsl-plugins/output",
+                "build/groovy-dsl-plugins/output"
             })
     void format_ignores_excluded_directories(String srcDir, GradleInvoker gradle, RootProject project) {
         project.buildGradle().append("""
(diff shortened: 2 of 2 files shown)


gradle-palantir-java-format/src/test/java/com/palantir/javaformat/gradle/SpotlessExcludesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @ParameterizedTest
+    @ValueSource(
+            strings = {
+                "build/generated/java",
+                "src/generated/java",
+                "src/main/generated_testsrc",
+                "generated_testSrc",
+                "build/groovy-dsl-plugins/output",
```
kelvinou01: can we also test some edge cases which should be included, like `src/foo/generated`

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
```diff
@@ -54,14 +54,15 @@ public void execute(JavaExtension java) {
...
-        java.targetExclude("**/groovy-dsl-plugins/**");
+        java.targetExclude(
+                "**/build/generated*/**",
+                "**/src/generated*/**",
+                "**/generated_*src/**",
+                "**/generated_*Src/**",
+                // build/groovy-dsl-plugins contains Java wrapper classes for
+                // https://docs.gradle.org/9.3.1/userguide/implementing_gradle_plugins_convention.html
+                // and don't need to be formatted.
+                "**/groovy-dsl-plugins/**");
```
kelvinou01, line 65: I'll open a PR to suppressible-error-prone aligning with these.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
```diff
@@ -54,14 +54,15 @@ public void execute(JavaExtension java) {
...
-        java.targetExclude("**/generated_*Src/**");
-        // build/groovy-dsl-plugins contains Java wrapper classes for
-        // https://docs.gradle.org/9.3.1/userguide/implementing_gradle_plugins_convention.html
-        // and don't need to be formatted.
-        java.targetExclude("**/groovy-dsl-plugins/**");
+        java.targetExclude(
+                "**/build/generated*/**",
+                "**/src/generated*/**",
+                "**/generated_*src/**",
+                "**/generated_*Src/**",
```
kelvinou01, line 61: small correctness thing which might not matter: will this match `generated_foo_src`? 

the current behavior [matches SEP](https://github.com/palantir/suppressible-error-prone/blob/611ff7acd4462afd6aa0464ea6925c0190bd556f/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L223)

crogoz: yes, it does - I used the same logic as in error prone: https://github.com/palantir/suppressible-error-prone/blob/develop/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L225.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
crogoz: they should be aligned already, no ? https://github.com/palantir/suppressible-error-prone/blob/0f2c42970e20b98d1b569e15ca9e86afc1c52a66/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L225

except for the groovy-dsl-plugins addition - which I am not really sure what it is and wether it is relevant to error prones

gradle-palantir-java-format/src/test/java/com/palantir/javaformat/gradle/SpotlessExcludesTest.java
crogoz: updated [here](https://github.com/palantir/palantir-java-format/pull/1690/commits/8c5f3819fc0fade48ead54d4fdde0906e8f7f364)

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
kelvinou01: if that's the case, let's explicate this in a test case? to document this behavior + make it easy to include  generated_foo_src for formatting in the future if we want to .

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
kelvinou01: yeah it seems to be the same. 

we false positive on a few cases in both repos, but that is a separate change!

```
foo/generated_foo/blah/src/bar
foo/build/generatedbar/baz
foo/src/generatedxforms/bar
```

gradle-palantir-java-format/src/test/java/com/palantir/javaformat/gradle/SpotlessExcludesTest.java
```diff
@@ -0,0 +1,102 @@
...
+            public class Test { Void test() { return null; } }
+            """);
+
+        InvocationResult result = gradle.withArgs("spotlessJavaCheck").buildsSuccessfully();
+
+        assertThat(result).task(":spotlessJava").succeeded();
+    }
+
+    @Test
+    void format_checks_non_generated_files(GradleInvoker gradle, RootProject project) {
```
kelvinou01: Lets make this test parameterized as well? It'd be nice to quickly grok what's included and excluded by glancing at the two tests. 

Let's add a few non-obvious test cases? It'd be good to explicate this behavior for our future selves.

```
foo/generated_foo_src/bar
foo/generated_foo/blah/src/bar
foo/build/generatedbar/baz
foo/src/generatedxforms/bar
```

crogoz: added test cases [here](https://github.com/palantir/palantir-java-format/pull/1690/commits/85282cf2357c16335e1b92981b0a8fbca8483981). Fysa - from the above list only the `foo/generated_foo/blah/src/bar` will be not-ignored

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
crogoz: I think line 54, 56 are explaining that we exclude generated source directories. This is where people usually set up these generated source sets.

gradle-palantir-java-format/src/main/java/com/palantir/javaformat/gradle/SpotlessInterop.java
kelvinou01: from internal discussion, ^ these aren't false positives.

kelvinou01 (approved):

PR #1731, opened 2026-08-01
feat(map-format): format key-value pairs to be inline

### LINKED ISSUE #1602: Make Map.of() and ImmutableMap.of() more readable
Currently, the formatter will format Map.of() as either:

```
// Shorter list
Map<String, String> map1 = Map.of("key1", value1, "key2", value2, "key3", value3);

// Longer list
Map<String, String> map2 = Map.of(
        "key1",
        longerValue1,
        "key2",
        longerValue2,
        "key3",
        longerValue3,
        "key4",
        longerValue4,
        "key5",
        longerValue5,
        "key6",
        longerValue6);
```

These are not very readable, especially when the keys are not visually distinguishable from values.

It would be much more readable if it would format them like this instead:

```
// Shorter list
Map<String, String> map1 = Map.of(
        "key1", value1,
        "key2", value2,
        "key3", value3);

// Longer list
Map<String, String> map2 = Map.of(
        "key1", longerValue1,
        "key2", longerValue2,
        "key3", longerValue3,
        "key4", longerValue4,
        "key5", longerValue5,
        "key6", longerValue6);
```

You may need to special case Map.of() and ImmutableMap.of(), as I don't think there is any more general rule.

ranpy13:
## Before this PR
the static factory methods for Map and ImmutableMaps are treated as regular methods, where each argument is expected to be in a new line. But for these methods, it makes more sense to have the key-value pairs inline, for better readability and context.

## After this PR
the key-value pairs of Map factory method will be in the same line, allowing for better readability.

## Possible downsides?
previously correctly formatted files will now show up as failing formatting checks and can potentially create a big diff across a lot of files. 

closes: https://github.com/palantir/palantir-java-format/issues/1602

### DIFF
--- a/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
+++ b/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java
@@ -3269,9 +3269,14 @@ public class JavaInputAstVisitor extends TreePathScanner<Void, Void> {
         return MAP_FACTORY_TYPES.contains(receiverName);
     }
 
-    private static final ImmutableSet<String> MAP_FACTORY_TYPES =
-            ImmutableSet.of("Map", "ImmutableMap", "ImmutableSortedMap", "ImmutableBiMap", "SortedMap", "NavigableMap",
-                    "ConcurrentMap");
+    private static final ImmutableSet<String> MAP_FACTORY_TYPES =ImmutableSet.of(
+            "Map",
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/palantir-java-format](https://github.com/palantir/palantir-java-format), @ranpy13! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1778, opened 2026-09-15
Don't remove java.lang imports that disambiguate a simple name

### LINKED ISSUE #1334: Incorrectly removes java.lang import that disambiguates an import or same-package class
Version: 2.68.0

Consider the following set up
```java
// example/src/main/java/com/company/project/foo/Byte.java
package com.company.project.foo;

public class Byte {

}

// example/src/main/java/com/company/project/bar/Test.java
package com.company.project.bar;

import com.company.project.foo.*;

import java.lang.Byte;

public class Test {
    public Byte foo() {
        // ...
    }
}
```

Assume there are other relevant classes under `src/main/java/com/company/project/foo`.

When running palantir-java-format on `Test.java`, the `java.lang.Byte` is removed. However, this is incorrect, as it was used to disambiguate `com.company.project.foo.Bar`. This ends up leading to a compilation failure as we use the wrong `Byte` class.

---

Or more generally, if `Byte` was in the same package, and there was no other import:

```java
// example/src/main/java/com/company/project/bar/Byte.java
package com.company.project.bar;

public class Byte {

}

// example/src/main/java/com/company/project/bar/Test.java
package com.company.project.bar;

import java.lang.Byte;

public class Test {
    public Byte foo() {
        // ...
    }
}
```
The `java.lang.Byte` import is still removed. This one is not as easy to workaround compared to the former (where we could've just expanded the wildcard imports)

ZayanKhan-12:
Fixes #1334.

## Before this PR

`RemoveUnusedImports` treated *every* import whose qualifier is `java.lang` as unused and deleted it unconditionally:

```java
if (qualifier.equals("java.lang")) {
    return true;
}
```

That is not sound. Per [JLS 6.4.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-6.html#jls-6.4.1), a
single-type-import declaration shadows both *types declared elsewhere in the current package* and *types brought in by
an import-on-demand*. So `import java.lang.Byte;` can be the only thing making a file compile, and removing it breaks
the build.

Both cases from the issue reproduce against `javac` (Corretto 21):

**1. Disambiguating an on-demand import**

```java
package com.company.project.bar;
import com.company.project.foo.*;   // also declares a `Byte`
import java.lang.Byte;
public class Test { Byte foo() { ... } }
```

Removing the import:

```
error: reference to Byte is ambiguous
  both class com.company.project.foo.Byte in com.company.project.foo
  and class java.lang.Byte in java.lang match
```

**2. Disambiguating a class in the same package**

```java
package com.company.project.bar;    // also contains Byte.java
import java.lang.Byte;
public class Test { Byte foo() { ... } }
```

Removing the import silently rebinds `Byte` to `com.company.project.bar.Byte`:

```
error: cannot find symbol
  symbol:   method valueOf(byte)
  location: class Byte
```

This is the worse of the two, because it is not a workaround away — there is no wildcard import to expand.

## After this PR

A `java.lang` **single-type** import is now only removed when its simple name is not referenced anywhere in the file
(code or javadoc) — i.e. when it is genuinely unused.

`RemoveUnusedImports` sees a single compilation unit with no classpath and no symbol resolution, so it cannot prove
that no clashing type exists in the current package. Where it cannot prove the edit is semantics-preserving, it now
declines to make it. A formatter silently changing which class a name binds to is a much worse failure than leaving a
redundant import in place.

Unchanged behaviour:

- `import java.lang.*;` is still removed — `java.lang` is implicitly imported on demand, so it is always redundant.
- Unused `java.lang.Foo` imports are still removed.
- Nested types such as `import java.lang.Foo.Bar;` were never on this code path (their qualifier is `java.lang.Foo`).
- Imports of types in the current package are still removed. Verified this is safe: package members already shadow
  import-on-demand declarations, so a same-package single-type import can never be disambiguating anything.

The only intentional behaviour change is that a **used** `java.lang.X` import is now preserved rather than stripped.
One existing test case encoded the old behaviour and has been updated with a comment explaining why.

While in here I also pulled the `*`-check and the used-name check into named helpers (`isOnDemand`, `isUsed`) so the
three branches read uniformly.

## Testing

- Five new cases in `RemoveUnusedImportsTest` covering: both scenarios from the issue, javadoc-only usage,
  `import java.lang.*;`, and an unused `java.lang` import alongside a wildcard import.
- Both issue scenarios were verified end to end against `javac` before and after the change.
- `./gradlew build` passes.

## Also included

A `CLAUDE.md` at the repo root documenting build/test commands, the module map, the formatter pipeline, the
file-based golden-test workflow, and the single-file-analysis constraint that this bug is an instance of — so that
agent-assisted contributions start from accurate context.


palantirtech:

PR #1781, opened 2026-09-15
Join string concatenations that end up on the same line

### LINKED ISSUE #68: Join string concatenations if they end up on the same line
## What happened?

```java
throw new SafeIllegalArgumentException("Invalid recipient field - contains multiple "
        + "foreign key properties.", UnsafeArg.of("field", maybeField));
```

gets turned into

```java
throw new SafeIllegalArgumentException(
        "Invalid recipient field - contains multiple " + "foreign key properties.", UnsafeArg.of("field", maybeField));
```

but the `+` is left behind.

There should be a single string literal and no `+`.

Implementation wise, my intuition is that we should rewrap string concatenations that end up being moved in normal formatting, which will achieve this.
Currently, `StringWrapper.wrap` will bail out if there are no overly long lines.

ZayanKhan-12:
Fixes #68.

## The problem

When normal formatting brings the operands of a string concatenation back onto one line, the `+` is left behind:

```java
throw new SafeIllegalArgumentException(
        "Invalid recipient field - contains multiple " + "foreign key properties.",
        UnsafeArg.of("field", maybeField));
```

The issue diagnoses this correctly: `StringWrapper` only reflows when a line exceeds the column limit, and a concatenation that now fits leaves no over-long line, so `needWrapping` short-circuits and nothing removes the `+`.

## The fix

A new pass in `Reflower` replaces each maximal run of adjacent string literals **that already share a line** with the single literal it is equivalent to. Both fast paths (`needWrapping`, and `linesNeedWrapping` used by the IntelliJ plugin) now also treat a joinable line as work to do — their javadoc asks that they be kept in line with each other, so both were updated.

The pass runs last and skips any range the wrapping or text-block passes already claimed, so it can't fight them.

After:

```java
throw new SafeIllegalArgumentException(
        "Invalid recipient field - contains multiple foreign key properties.",
        UnsafeArg.of("field", maybeField));
```

## What is deliberately *not* joined

| Case | Why |
| --- | --- |
| Literals still split across lines | That's deliberate wrapping; `wrapLongStrings` owns it |
| A run separated by a comment | Keeps `//$NON-NLS-1$` markers attached to their literals |
| Text blocks | `indentTextBlocks` owns those |
| `"\1" + "2"` | `\1` then `2` is **not** `\12` — see below |

The escape case is the subtle one. Java octal escapes are up to three digits, so splicing a digit onto a shorter one silently changes the character. `joiningWouldChangeEscapes` refuses those, while still joining `"\377" + "7"` (already maximal) and `"back\\" + "1"` (even backslash run, so the digits are literal text).

Joining only ever *shortens* a line — it removes ` + ` and two quotes — so it cannot push a line over the limit or oscillate with the wrapping pass.

## Why I'm confident this is safe

`StringWrapper.wrap` already re-parses with `allowStringFolding=true` and asserts the AST is unchanged, which is exactly the right net for this change: a joined literal and the split form fold to the same constant.

Beyond that:

- **Differential execution test.** I wrote a program exercising octal escapes, escaped backslashes, tabs, quotes, newlines and non-literal operands, ran it, formatted it, and ran it again. Output was **byte-identical** (same md5). The `"\1" + "2"` case survived unjoined, as intended.
- **Corpus run.** Formatted 137 of this repo's own source files with the patched build: no crashes, no AST mismatches. 7 files had concatenations joined, and every one was a genuine leftover `+` — e.g. `"Replacement{" + "replaceRange="` in `Replacement.java`. That run also caught that my own new code wasn't formatted to project style, which I then fixed.
- **Idempotency** is covered by `FormatterIntegrationTest`'s existing `idempotentLF`/`idempotentCR` templates, and verified by hand by formatting twice.

## Changed expectations

Exactly two existing goldens change, both to the behaviour #68 asks for:

- `NON-NLS`: the uncommented constant joins to `"field_0,field_1,field_2,field_3,field_4"`. **Its `//$NON-NLS-1$`-commented sibling is untouched** — that was the case I most wanted to get right, since those markers must stay with their strings.
- `palantir-10`: `foo("foo" + "bar" + THING + "baz")` becomes `foo("foobar" + THING + "baz")`. Its input still spans two lines, so it continues to exercise the reflow it was written for; the multi-line `@SqlUpdate` concatenation above it is correctly left alone.

Nothing else in the 452 testdata files moved.

## Testing

- New golden `palantir-join-string-concatenations` covering the issue's case plus every exclusion above.
- Unit tests for `joiningWouldChangeEscapes` and `mayHaveJoinableLiterals`.
- `./gradlew build` passes: 1412 tests, 0 failures.

## Note


palantirtech:

Issue #215, opened 2020-02-21
Allow Inlined Annotations for Mocks


gm2211:
## What happened?
```
 @Mock(answer = Answers.RETURNS_DEEP_STUBS) private Job job;
```

gets reformatted as 

```
@Mock(answer = Answers.RETURNS_DEEP_STUBS) 
private Job job;
```

Which is not very space-efficient if you have a bunch of mocks

## What did you want to happen?

I'd like the code to stay with the inlined annotation.

dansanduleac:
The main goal of the formatter is not space efficiency, but readability.
Even if we can't possibly write heuristics that always produce the best readability im all cases, that is the goal we're moving towards, and we'll gladly make things more space efficient by inlining _if_ it improves visibility. Or rather, if the alternative hurts readability.

In this case though, I think inlining hurts readability, because it's now harder to quickly scan variable types/names by just looking at the left column, as your eyes have to scan right every time, and the column where they get defined will vary too with the length of each annotation.

Furthermore, when having annotated fields like this, PJF currently inserts a space in between them too, to make it even easier to scan.

```java
@Mock(answer = Answers.RETURNS_DEEP_STUBS) private Foo foo;
@Mock(answer = Answers.RETURNS_DEEP_STUBS) private Bar bar;
```

will get reformatted as this, which I would argue has a lot more readability:

```java
@Mock(answer = Answers.RETURNS_DEEP_STUBS)
private Foo foo;

@Mock(answer = Answers.RETURNS_DEEP_STUBS)
private Bar bar;
```

gm2211:
To be honest, I don't really care about space-efficiency per-se, in fact people complain about Java being verbose, but I have no issues with that.

What I do care about is when visual clutter makes things harder to navigate or overwhelms the reader.

I guess this is very subjective, but I disagree that the latter option you propose is any more readable, especially in classes that have many mocks.

Also, I probably picked the wrong example by using a mock annotation with parameters, but consider this:


```java
private static final SomeType SOME = SomeType.builder()
     ...
    .build();
private static final SomeType2 SOME_2 = SomeType2.builder()
     ...
    .build();
private static final SomeType3 SOME_3 = SomeType3.builder()
     ...
    .build();
private static final SomeType4 SOME_4 = SomeType4.builder()
     ...
    .build();

@Mock private SomeObjMetadata meta;
@Mock private SomeContainerObject containerObj;
@Mock private Foo foo;
@Mock private Bar bar;
[...]

PR #1782, opened 2026-09-15
Keep a single marker annotation on the field's line

ZayanKhan-12:
Proposes a fix for #215. **This one is a deliberate style change, not a bug fix — it needs a maintainer's taste call, and I've tried to make it easy to say no to.** Details below.

## Current behaviour

```java
@Mock
private Foo foo;

@Mock
private Bar bar;
```

Fields are the only declaration kind hardcoded to vertical annotations:

```java
visitVariables(
        variableFragments(it, bodyDeclaration),
        DeclarationKind.FIELD,
        // We always want field annotations to be vertical
        Direction.VERTICAL);
```

## The change

Pass the predicate that locals and parameters already use, `canLocalHaveHorizontalAnnotations` (renamed here to `canHaveHorizontalAnnotations`, since it no longer serves only locals). It returns `HORIZONTAL` only for **at most one argument-less annotation**:

```java
@Mock private Foo foo;
@Mock private Bar bar;
```

The supporting machinery already exists and is already documented in `declareOne`:

```java
// If the node is a field declaration, try to output any declaration
// annotations in-line. If the entire declaration doesn't fit on a single
// line, fall back to one-per-line.
```

So the blank line is already conditional on the vertical break happening (`BlankLineWanted.conditional(verticalAnnotationBreak)`), and a declaration too wide to inline still falls back to one-per-line. I verified that fallback with a field that overflows 120 columns.

## Explicitly unchanged

`@dansanduleac`'s objection on the issue was that inlining makes the left column harder to scan, because the type's column varies with the annotation's length. That argument is strongest exactly where this change does **not** apply:

| Stays vertical | Example |
| --- | --- |
| Annotations with arguments | `@Mock(answer = Answers.RETURNS_DEEP_STUBS)` |
| More than one annotation | `@Mock @Inject` |
| Anything that overflows the column limit | verified in the new test |
| Methods and types | `@Override`, `@Deprecated` on a class |

A single marker annotation shifts the type by a short, fixed amount. An annotation with arguments shifts it an arbitrary, varying distance — which is the case the objection describes, and the one the reporter later said they'd picked by mistake ("I probably picked the wrong example by using a mock annotation with parameters").

The narrowed ask — `@Mock private Foo foo;` — was never replied to on the issue.

## Why I think it's defensible

Less about space efficiency, more about consistency: palantir-java-format **already** inlines a single marker annotation on locals and parameters. Fields being the sole exception is the part that's hard to justify, and it's what makes a class of mocks read very differently from the same annotations elsewhere.

## Blast radius

Small and entirely of the targeted shape. Four goldens plus one hardcoded expectation in `FormatterTest`:

```diff
-    @Nullable
-    public int f;
+    @Nullable public int f;
```

Nothing else among the 452 testdata files moved. All idempotency checks pass.

This will still reformat every `@Nullable`/`@Inject`/`@Mock` field in every downstream codebase, which is exactly why it's your call and not mine.

## Testing

- New golden `palantir-inline-marker-annotation-fields` covering each case in the table above plus the overflow fallback and a local, so the boundary is pinned.
- `./gradlew build` passes: 1409 tests, 0 failures.

## If you'd rather not

Happy to close this, or to rework it as an opt-in rather than a default. I'd rather leave the formatter opinionated than add a knob, so I didn't propose one — but say the word.

## Note


palantirtech:

PR #1783, opened 2026-09-15
Don't require argument columns after the first to be parallel

### LINKED ISSUE #204: Two-column argument layout breaks with different AST shapes
We've hit an obscure edge case for the two-column argList logic, where it doesn't end up keeping the pairs together.
The logic tries to assert that N/2+1 rows have the same "kind" of tree in the AST.
However, the values in the example seem to be 
* Kind.PLUS (BinaryTree of `+` operations)
* Kind.VARIABLE
* Kind.METHOD_INVOCATION

https://github.com/palantir/palantir-java-format/blob/fc94362438d9450b2fefe29315e5ea687bd93215/palantir-java-format/src/main/java/com/palantir/javaformat/java/JavaInputAstVisitor.java#L3222-L3226

I honestly don't see much reason for this kind of heuristic (at least for values), so it might make sense to remove it.

_Originally posted by @dansanduleac in https://github.com/palantir/conjure-verification/pull/311#discussion_r377215499_

ZayanKhan-12:
Fixes #204, taking the fix direction suggested in the issue:

> I honestly don't see much reason for this kind of heuristic (at least for values), so it might make sense to remove it.

## The bug

`argumentsAreTabular` groups arguments into rows by the column they start on **in the source**, so it only ever fires on a grid the author has already laid out. It then required every column to be "parallel":

```java
if (!expressionsAreParallel(rows, 0, rows.size())) {      // column 0: all rows the same Kind
    return -1;
}
for (int i = 1; i < size0; i++) {
    if (!expressionsAreParallel(rows, i, rows.size() / 2 + 1)) {   // rest: a majority
        return -1;
    }
}
```

That second condition discards the author's grid based on something that says nothing about how the arguments read. A values column holding a `PLUS`, an identifier and a method invocation has no majority kind, so this:

```java
assertThat(
        "alpha", first + second,
        "bravo", someVariable,
        "charlie", compute(),
        "delta", other + thing);
```

collapsed onto a single line, while the structurally identical call below kept its table purely because its values happened to be uniform:

```java
assertThat(
        "alpha", computeA(),
        "bravo", computeB(),
        ...
```

## The fix

Drop the check for columns after the first.

The leading column must **still** be uniform — that's what makes a set of arguments recognisable as a table — and every row-length and start-column constraint is untouched. So this only ever *keeps* a grid the author wrote and column 0 already vouched for; it never invents one.

## The diff speaks for itself

The only existing expectation that changes is `PairedArguments`, the golden written specifically to exercise these heuristics. It already contained two adjacent, nearly identical calls that formatted completely differently:

```diff
         f(
-                xxxxxxxxxxxxxxxxxxxxxxxxx,
-                xxxxxxxxxxxxxxxxxxxxxxx(),
-                xxxxxxxxxxxxxxxxxxxxxxxxx,
-                xxxxxxxxxxxxxxxxxxxxxxx(),
-                xxxxxxxxxxxxxxxxxxxxxxxxx,
-                xxxxxxxxxxxxxxxxxxxxxxxxx,
-                xxxxxxxxxxxxxxxxxxxxxxxxx,
-                xxxxxxxxxxxxxxxxxxxxxxxxx);
+                xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxx(),
+                xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxx(),
+                xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxxxx,
+                xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxxxx);
         f(
                 xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxx(),
                 xxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxxxx,
```

Those two calls differ only in whether the values column has two method calls or one — 2 of 4 rows versus 3 of 4, either side of the `rows.size() / 2 + 1` threshold. After this change they agree. That inconsistency is the bug, stated as a test.

Nothing else among the 452 testdata files moved.

## Testing

New golden `palantir-issue-204-tabular-mixed-values` pinning all four behaviours, each verified against 2.98.0 to confirm it actually demonstrates something:

| Case | Before | After |
| --- | --- | --- |
| Two-column call, mixed value kinds | one arg per line | **table preserved** |
| Array initialiser, 3 columns, mixed kinds | one element per line | **grid preserved** |
| Leading column *not* uniform | one arg per line | unchanged — guard still bites |
| Uniform values | table | unchanged |

Two things worth calling out from writing that test:

- My first draft of the array-initialiser case was **already** tabular before the fix (its columns happened to have majority kinds), so it proved nothing. I reworked it until the columns had all-distinct kinds and confirmed the before/after differ.
- The four-column *method argument* case is deliberately included and stays non-tabular: `addArguments` only applies the grid when `argumentsAreTabular(...) == 2`, so method arguments are two-column only regardless. Array initialisers are the caller that uses the N-column result. The test documents that split.

`./gradlew build` passes: 1409 tests, 0 failures. Idempotency verified by hand and by the existing `idempotentLF`/`idempotentCR` templates.

## Note


palantirtech:

PR #1337, opened 2025-06-29
fix `RemoveUnusedImportsStep` leftovers

Pankraz76:
depends on:
- https://github.com/diffplug/spotless/pull/2541

ref:
- https://github.com/google/google-java-format/pull/1262

fix #1261

implementation fixed by reliable contributor @iddeepak

Thanks for this initiative.

- https://github.com/google/google-java-format/pull/1262
- https://github.com/google/google-java-format/pull/1262#issuecomment-3014695342

before: 14 issues:

image

now: 10 issues:

image

PR #1784, opened 2026-09-15
Run the string wrapping pass to a fixed point

### LINKED ISSUE #1343: string block collaps on empty line
when removing `// tmp fix` comment the block collapses like seen below:

```
            {
                {
                    """
                    import java.util.List;
                    import java.lang.Deprecated;

                    @Deprecated
                    class Test {}
                    """
                },
                // fixme:
                //  this is a bug in formatter
                //  when line is not blank but having white spaced block remains stable
                {
                    """
                    // tmp fix 
                    @Deprecated
                    class Test {}
                    """
                },
                {"""

                @Deprecated
                class Test {}
                """}
            },
```

- https://github.com/palantir/palantir-java-format/pull/1337/files#r2182497166

ZayanKhan-12:
Fixes #1343.

## What the issue reports, and what's actually wrong

The issue reports a text block "collapsing" when its first line is blank. Chasing it, the underlying defect is sharper and more serious than that: **the formatter is not idempotent here.** `format(format(x)) != format(x)`.

Formatting the file from the linked PR (#1337) twice moves **582 lines** on the second pass. A file palantir-java-format has just written is still reported as unformatted by `--set-exit-if-changed` and by spotless — so a CI format check fails on output the formatter itself produced.

Reduced to three text blocks' worth of source:

```java
String[][] x = {
    {
        """
        // tmp fix
        @Deprecated
        class Test {}
        """
    },
    {
        """

        @Deprecated
        class Test {}
        """
    }
};
```

The second text block's content lands at column 16 on the first pass and column 12 on the second. That's also why deleting the `// tmp fix` line appears to "collapse" the block: it changes which round the layout settles on.

## Cause

`StringWrapper` derives a text block's indentation from the layout around it (`computeCustomTextBlocksIndent` / the default prefix in `indentTextBlocks`, both reading source positions). But the same pass moves that layout, so the prefix it computed can already be stale by the time it is applied.

`wrap` was clearly aware of this — it already recomputes its replacements after reformatting, **twice**:

```java
if (!firstPass.equals(input)) {
    input = firstPass;
    replacements = getReflowReplacements(columnLimit, input);
}
...
if (!secondPass.equals(result)) {
    replacements = getReflowReplacements(columnLimit, secondPass);
    result = applyReplacements(secondPass, replacements);
}
```

It just stops at two rounds, and nested array initialisers need three.

## The fix

Extract the existing body as `wrapOnce` and iterate it while it keeps changing the source, bounded by `MAX_ROUNDS = 5` so a pathological input can't spin. That's the existing design taken to its conclusion rather than a new mechanism.

## Why this is low risk

- **One pass now lands on exactly the result the old code reached after two.** I verified `format_new(x) == format_old(format_old(x))` byte-for-byte on the 1034-line file from #1337. The fix doesn't invent a new layout; it just converges immediately to the one the old code would have reached anyway.
- **No existing expectation changes.** All 452 testdata goldens regenerate identically — for every existing test input, one round was already a fixed point.
- **No measurable cost.** The loop body only re-runs when the previous round actually changed the source, so unchanged files pay nothing. Across 137 of this repo's own files: ~1.17s before, ~1.16s after (3 runs each) — within noise.
- The `MAX_ROUNDS` bound means the worst case is the current behaviour, not a hang.

## Testing

New golden `palantir-issue-1343-text-block-indent-stable`, covering a blank first line and a blank line mid-block, both inside nested array initialisers.

`FormatterIntegrationTest` runs every golden through `idempotentLF` and `idempotentCR`, which is exactly the invariant that was broken — so this test is meaningful rather than decorative. I confirmed that by stashing the fix and re-running:

```
FormatterIntegrationTest > format() > palantir-issue-1343-text-block-indent-stable FAILED
908 tests completed, 1 failed
```

With the fix, `./gradlew build` passes: 1409 tests, 0 failures.

## Note

This fixes the idempotency violation. If there is also a *preferred* indentation for a text block that opens with a blank line, that's a separate judgement call about which fixed point is nicest, and I'd rather settle the correctness bug first. Happy to follow up if you have a preference.


palantirtech:

